PR #4, opened 2024-12-02
Add code action to implement missing trait members

integraledelebesgue:
Closes  [starkware-libs/cairo#5944](https://github.com/starkware-libs/cairo/issues/5944#issue-2386254154)

## Changes
* If `impl` block is missing some members of an implemented trait, a quick fix filling those members appears
* Known generic arguments are inserted in impls, unknown substituted with placeholders
* Inserted code is not formatted

### DIFF
--- a/src/ide/code_actions/fill_trait_members.rs
+++ b/src/ide/code_actions/fill_trait_members.rs
@@ -1,0 +1,285 @@
+use std::collections::HashMap;
+
+use cairo_lang_defs::ids::{NamedLanguageElementId, TraitConstantId, TraitFunctionId};
+use cairo_lang_semantic::db::SemanticGroup;
+use cairo_lang_semantic::diagnostic::{NotFoundItemType, SemanticDiagnostics};
+use cairo_lang_semantic::expr::inference::InferenceId;
+use cairo_lang_semantic::lookup_item::HasResolverData;
+use cairo_lang_semantic::resolve::{ResolvedConcreteItem, Resolver};
+use cairo_lang_semantic::substitution::{
+    GenericSubstitution, SemanticRewriter, SubstitutionRewriter,
+};
+use cairo_lang_semantic::{ConcreteTraitId, GenericArgumentId, GenericParam, Parameter};
+use cairo_lang_syntax::node::ast::{ImplItem, ItemImpl, MaybeImplBody};
+use cairo_lang_syntax::node::kind::SyntaxKind;
+use cairo_lang_syntax::node::{SyntaxNode, Token, TypedSyntaxNode};
+use itertools::{Itertools, chain};
+use lsp_types::{CodeAction, CodeActionKind, CodeActionParams, Range, TextEdit, WorkspaceEdit};
+
+use crate::lang::db::{AnalysisDatabase, LsSemanticGroup, LsSyntaxGroup};
+use crate::lang::lsp::ToLsp;
+
+/// Generates a completion adding all trait members that have not yet been specified.
+/// Functions are added with empty bodies, consts with placeholder values.
+pub fn fill_trait_members(
+    db: &AnalysisDatabase,
+    node: SyntaxNode,
+    params: &CodeActionParams,
+) -> Option<CodeAction> {
+    let file = db.find_module_file_containing_node(&node)?.file_id(db).ok()?;
+
+    let item_impl_node = db.first_ancestor_of_kind(node, SyntaxKind::ItemImpl)?;
+    let item_impl = ItemImpl::from_syntax_node(db, item_impl_node);
+
+    // Do not complete `impl`s without braces.
+    let MaybeImplBody::Some(impl_body) = item_impl.body(db) else {
+        return None;
+    };
+
+    let specified_impl_items = impl_body.items(db);
+
+    let already_implemented_item_names = specified_impl_items
+        .elements(db)
+        .iter()
+        .filter_map(|item| match item {
+            ImplItem::Function(item) => Some(item.declaration(db).name(db).token(db).text(db)),
+            ImplItem::Type(item) => Some(item.name(db).token(db).text(db)),
+            ImplItem::Constant(item) => Some(item.name(db).token(db).text(db)),
+            _ => None, // No other items can appear in trait impl.
+        })
+        .collect_vec();
+
+    let concrete_trait_id = find_concrete_trait_id(db, &item_impl)?;
+    let trait_id = concrete_trait_id.trait_id(db);
+
+    let mut trait_constants = db.trait_constants(trait_id).ok()?;
+    let mut trait_types = db.trait_types(trait_id).ok()?;
+    let mut trait_functions = db.trait_functions(trait_id).ok()?;
+
+    trait_constants.retain(|key, _| !already_implemented_item_names.contains(key));
+    trait_types.retain(|key, _| !already_implemented_item_names.contains(key));
+    trait_functions.retain(|key, _| !already_implemented_item_names.contains(key));
+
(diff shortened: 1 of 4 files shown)


tests/test_data/code_actions/fill_trait_members.txt
```diff
@@ -0,0 +1,146 @@
...
+
+fn generic<const V: core::integer::u32, W, +Into<core::integer::u32, W>>(w: W) {}
+
+fn with_concrete_impl<W, impl SomeImpl: Into<core::integer::u32, W>>(w: W) -> W {}"
+At: Range { start: Position { line: 31, character: 47 }, end: Position { line: 31, character: 47 } }
+
+//! > Code action #3
+    impl ImplWithEverything<caret> of MyTrait<u32, felt252, 12> {
+Title: Implement missing members
+Add new text: ""
```
integraledelebesgue, line 137: @piotmag769 An empty string shouldn't appear here, I'm going to fix this.

src/ide/code_actions/fill_trait_members.rs
```diff
@@ -0,0 +1,278 @@
...
+/// Generates a completion adding all trait members that have not yet been specified.
+/// Functions are added with empty bodies, consts with placeholder values.
+pub fn fill_trait_members(
+    db: &AnalysisDatabase,
+    node: SyntaxNode,
+    params: &CodeActionParams,
+) -> Option<CodeAction> {
+    let file = db.find_module_file_containing_node(&node)?.file_id(db).ok()?;
+
+    let item_impl_node = db.first_ancestor_of_kind(node, SyntaxKind::ItemImpl)?;
```
integraledelebesgue, line 31: @piotmag769 Note: Only `impl`'s name is affected by the diagnostics and underlined in the `IDE`. Therefore, code action can be invoked only on the `impl` keyword, it's not available inside the body.

integraledelebesgue:
Note: this action will not work until changes to the compiler from [this PR](https://reviewable.io/reviews/starkware-libs/cairo/6690) are not merged.

src/ide/code_actions/fill_trait_members.rs
```diff
@@ -0,0 +1,278 @@
...
+
+            Some(param.id.name(db).map_or_else(
+                // concrete trait used only as a constraint
+                || format!("+{trait_name}{generic_arguments_bracket}"),
+                // concrete trait with explicit impl
+                |name| format!("impl {name}: {trait_name}{generic_arguments_bracket}"),
+            ))
+        }
+
+        other => Some(other.as_arg(db).format(db)),
```
piotmag769, line 237: `NegImpl` is not allowed here no?

src/ide/code_actions/fill_trait_members.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    let nopanic = if !signature.panicable { Some(String::from("nopanic")) } else { None };
+
+    let body: Option<String> = Some(String::from("{}"));
+
+    Some([title, return_type, implicits, nopanic, body].into_iter().flatten().join(" "))
+}
+
+/// Formats [`GenericParam`] to be used as a free-standing type parameter
+/// (not belonging to the trait) in a generic function's declaration.
```
piotmag769, line 201: That's not always true, sometimes you substitute trait generic types with concrete types, no?

integraledelebesgue: Indeed, docstring is inaccurate. This function uses substitution rewriter on impl-related parameters but keeps "free" generics unchanged, so it's functionally correct

tests/test_data/code_actions/fill_trait_members.txt
integraledelebesgue: Done :)

src/ide/code_actions/fill_trait_members.rs
piotmag769: Thanks for the explanation!

src/ide/code_actions/fill_trait_members.rs
```diff
@@ -0,0 +1,285 @@
...
+            let concrete_trait = param.concrete_trait.ok()?;
+            let trait_name = concrete_trait.name(db);
+            let trait_generic_arguments = concrete_trait.generic_args(db);
+
+            let generic_arguments_bracket = if trait_generic_arguments.is_empty() {
+                String::new()
+            } else {
+                let formatted_arguments = trait_generic_arguments
+                    .into_iter()
+                    .map(|argument| generic_argument_code(db, argument, type_concretizer))
```
piotmag769, line 227: [nit] If all generics were substituted with concrete traits here (e.g. `+Drop<T>` with `+Drop<felt252>`) cannot we just drop the trait contraint? Or does cairo not allow it? It is really a nitpick, you can create an issue for that

src/ide/code_actions/fill_trait_members.rs
```diff
@@ -0,0 +1,285 @@
...
+    let mut type_concretizer = SubstitutionRewriter { db, substitution: &substitution };
+
+    // Iterators borrowing SubstitutionRewriter mutably need intermediate collection.
+    let code = chain!(
+        trait_constants
+            .values()
+            .filter_map(|&id| constant_code(db, id, &mut type_concretizer))
+            .collect_vec()
+            .into_iter(),
+        trait_types.values().map(|id| format!("type {} = ();", id.name(db))),
```
Arcticae, line 79: I would put types above the constants i think

src/ide/code_actions/fill_trait_members.rs
```diff
@@ -0,0 +1,285 @@
...
+        .join(", ");
+
+    let name = id.name(db);
+    let title = Some(format!("fn {name}{generic_parameters_bracket}({parameters})"));
+
+    let return_type = type_concretizer.rewrite(signature.return_type).ok()?;
+    let return_type =
+        if return_type.is_unit(db) { None } else { Some(format!("-> {}", return_type.format(db))) };
+
+    let implicits = match &signature.implicits[..] {
```
Arcticae, line 192: do we need to generate those? 🤔

src/ide/code_actions/fill_trait_members.rs
Arcticae: It's safer to generate this more explicit version probably

src/ide/code_actions/mod.rs
```diff
@@ -88,6 +89,9 @@ fn get_code_actions_for_diagnostics(
             "E0003" => fill_struct_fields::fill_struct_fields(db, node.clone(), params)
                 .map(|result| vec![result])
                 .unwrap_or_default(),
+            "E0004" => fill_trait_members::fill_trait_members(db, node.clone(), params)
```
Arcticae, line 92: What happens if it doesn't return anything? Nothing happens? 
I think that might be weird UX for the users:

1. You click "fill trait members"
2. Nothing happens 
3. You wait, or re-run and still nothing

If it's the case, i think it should be handled in a different PR (create an issue please 🙏 ) since it looks like other actions follow this pattern too

src/ide/code_actions/fill_trait_members.rs
piotmag769: Yep, it is not allowed in cairo, just checked it. Resolving

src/ide/code_actions/mod.rs
piotmag769: >What happens if it doesn't return anything?

Then no code action is shown

```rs
                .map(|result| vec![result])
                .unwrap_or_default(),

Arcticae (approved):

PR #8, opened 2024-12-02
Refactor proc macro channels

Draggu:
**Stack**:
- #9
- #8 ⬅

### DIFF
--- a/src/lang/proc_macros/controller.rs
+++ b/src/lang/proc_macros/controller.rs
@@ -52,18 +52,12 @@ pub struct ProcMacroClientController {
     notifier: Notifier,
     plugin_suite: Option<PluginSuite>,
     initialization_retries: RateLimiter<NotKeyed, InMemoryState, QuantaClock>,
-    channels: Option<ProcMacroChannelsSenders>,
+    channels: ProcMacroChannels,
 }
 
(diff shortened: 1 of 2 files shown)


src/lang/proc_macros/controller.rs
```diff
@@ -256,19 +251,25 @@ fn parse_proc_macro_response(response: RpcResponse) -> Result<ProcMacroResult> {
 }
 
 #[derive(Clone)]
-pub struct ProcMacroChannelsReceivers {
+pub struct ProcMacroChannels {
```
piotmag769, line 254: Imo we should have separate structs for senders and receivers (can be substructs too)

Draggu: There is no need to do this, it would make it less clear only.

piotmag769: I disagree, it is a weird struct with 2 pubs, but I can give it a pass

mkaput: I'd agree with Piotr M, but indeed this looks like a style thing. Let's stay this as-is to not bike-shed.

mkaput (approved):

piotmag769 (approved):

PR #13, opened 2024-12-03
Explain invalidate of `priv_raw_file_content` as it is not necessary in current form

Draggu:
Since https://github.com/starkware-libs/cairo/pull/6799, we can use `cancel_all` as well.
We keep this code to make it clear what we aim to achieve here.

### DIFF
--- a/src/server/routing/traits.rs
+++ b/src/server/routing/traits.rs
@@ -185,6 +185,14 @@ impl SyncNotificationHandler for DidChangeWatchedFiles {
         for change in &params.changes {
             if is_cairo_file_path(&change.uri) {
                 let Some(file) = state.db.file_for_url(&change.uri) else { continue };
+                // Invalidating the `priv_raw_file_content` query to be
+                // recomputed, works similar to manually triggering a low-durability synthetic
+                // write (this is what
+                // [`crate::lang::db::AnalysisDatabase::cancel_all`] does) to refresh it.
+                //
+                // We opt for this approach instead of using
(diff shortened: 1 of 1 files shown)


piotmag769 (approved):

src/server/routing/traits.rs
```diff
@@ -181,14 +181,6 @@ impl SyncNotificationHandler for DidChangeWatchedFiles {
         requester: &mut Requester<'_>,
         params: DidChangeWatchedFilesParams,
     ) -> LSPResult<()> {
-        // Invalidate changed cairo files.
-        for change in &params.changes {
-            if is_cairo_file_path(&change.uri) {
-                let Some(file) = state.db.file_for_url(&change.uri) else { continue };
-                PrivRawFileContentQuery.in_db_mut(state.db.as_files_group_mut()).invalidate(&file);
```
mkaput, line 188: just to make sure: have you verified that this works?

Draggu: It works fine for me, but if you see any case where this can break, please point it here

mkaput: It is weird because these invalidations are required according to https://murek.dev/salsa-book-0.16/common_patterns/on_demand_inputs.html

mkaput: How have you tested this?

piotmag769: We do synthetic read in `priv_raw_file_content`, no? So if the file content has changed it will be invalidated

mkaput (approved):

piotmag769 (approved):

PR #17, opened 2024-12-03
Remove dependency on `rustc-hash`

mkaput:
There is no hard data on any perf benefits of using it at the single place we do, and fewer dependencies is always a good improvement.

### DIFF
--- a/src/server/client.rs
+++ b/src/server/client.rs
@@ -6,11 +6,11 @@
 // +-----------------------------------------------------+
 
 use std::any::TypeId;
+use std::collections::HashMap;
 
 use anyhow::Result;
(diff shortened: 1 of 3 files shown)


src/server/client.rs
```diff
@@ -46,7 +46,7 @@ impl Client<'_> {
             requester: Requester {
                 sender,
                 next_request_id: 1,
-                response_handlers: FxHashMap::default(),
+                response_handlers: HashMap::default(),
```
Draggu, line 49:
```suggestion
                response_handlers: Default::default(),
```

Draggu (changes requested):

PR #23, opened 2024-12-03
Add Cairo update check workflow

mkaput:


.github/workflows/cairo-update-check.yml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: software-mansion/setup-scarb@v1
+        with:
+          scarb-version: nightly
```
piotmag769, line 25: Are we sure scarb will release nightly on daily basis? Why is Scarb even here?

.github/workflows/cairo-update-check.yml
```diff
@@ -0,0 +1,49 @@
...
+      slack:
+        description: "Send Slack notification on failure"
+        type: boolean
+        default: true
+  schedule:
+    - cron: '0 0 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
```
piotmag769, line 15: Why do we need it?

.github/workflows/cairo-update-check.yml
```diff
@@ -0,0 +1,49 @@
...
+        run: |
+          CAIRO_REV=$(git ls-remote --refs "https://github.com/starkware-libs/cairo" main | awk '{print $1}')
+          echo "::notice::Checking Cairo commit: https://github.com/starkware-libs/cairo/commit/$CAIRO_REV"
+          cargo xtask set-cairo-version --rev "$CAIRO_REV"
+
+      - run: cargo test --profile=ci --no-fail-fast
+
+  notify_failed:
+    runs-on: ubuntu-latest
+    if: always() && !(inputs.slack) && needs.check.result == 'failure'
```
piotmag769, line 40: What does `always()` do? Why `!(inputs.slack)`?

.github/workflows/cairo-update-check.yml
mkaput: This mostly came by copy-pasting, but in general for this workflow always the last run represents the most fresh state so we will never be interested in old runs, so why not?

.github/workflows/cairo-update-check.yml
piotmag769: Yea but why the `!`?

.github/workflows/cairo-update-check.yml
mkaput: discussed offline

.github/workflows/cairo-update-check.yml
mkaput: 🤦🏻

piotmag769 (approved):

PR #32, opened 2024-12-05
Add debouncer for proc macros

piotmag769:


Draggu:
What is the difference in total time used to resolve macros with and without this debouncer?

mkaput (approved):

Draggu (approved):

PR #33, opened 2024-12-06
Use patch-based Cairo dependencies specification

mkaput:


xtask/src/set_cairo_version.rs
```diff
@@ -45,83 +46,193 @@ pub fn main(args: Args) -> Result<()> {
...
+
+            // Add local path reference if requested.
+            // For local path sources, Cargo is not looking for crates recursively therefore, we
+            // need to manually provide full paths to Cairo workspace member crates.
+            if let Some(path) = &args.spec.path {
+                dep.insert(
+                    "path",
+                    path.join("crates").join(dep_name).to_string_lossy().into_owned().into(),
+                );
+            }
```
integraledelebesgue, line 130: How about checking the path for validity? At the moment, we can pass an arbitrary string which will be written to the manifest before calling `cargo fetch`

mkaput: This is fine for me because the aforementioned `cargo fetch` will always be called and thus validate it.

Same thing applies to version numbers and git refs.

Draggu (approved):

xtask/src/set_cairo_version.rs
```diff
@@ -45,83 +46,193 @@ pub fn main(args: Args) -> Result<()> {
...
     }
+
+    Ok(())
+}
+
+fn find_unused_patches(cargo_lock: &DocumentMut) -> Option<Vec<String>> {
+    Some(
+        cargo_lock
+            .get("patch")?
+            .get("unused")?
```
piotmag769, line 232: I don't quite understand this one, where would it come from?

piotmag769 (approved):

xtask/src/set_cairo_version.rs
piotmag769: Nvm, I got it. What is the case in which it might happen though?

mkaput: Added docs answering your question

PR #35, opened 2024-12-11
Fix file batching

piotmag769:


Draggu (approved):

Arcticae (approved):

src/lang/diagnostics/mod.rs
```diff
@@ -111,10 +111,13 @@ impl DiagnosticsControllerThread {
     fn diagnostics_controller_tick(&self, state_snapshots: StateSnapshots) {
         let (state, primary_snapshots, secondary_snapshots) = state_snapshots.split();
 
-        let primary = find_primary_files(&state.db, &state.open_files);
+        let primary_set = find_primary_files(&state.db, &state.open_files);
+        let secondary = find_secondary_files(&state.db, &primary_set);
```
mkaput, line 115: why have you moved this line up? it was deliberately after spawning primary workers, so that they could begin they work before we start computing secondary files

mkaput (changes requested):

mkaput (approved):

PR #40, opened 2024-12-17
Add `MockClient::open_all_cairo_files_and_wait_for_project_update`

mkaput:


piotmag769 (approved):

tests/e2e/support/mock_client.rs
```diff
@@ -347,6 +351,26 @@ impl MockClient {
     pub fn wait_for_project_update(&mut self) {
         self.wait_for_notification::<ProjectUpdatingFinished>(|_| true);
     }
+
+    /// Sends `textDocument/didChange` notification to the server for each `*.cairo` file in test
+    /// fixture and then waits for all `cairo/projectUpdatingFinished` notifications.
```
Draggu, line 356: Does not wait for `all` `cairo/projectUpdatingFinished` but only for a corresponding number of those.

Draggu: I'm not sure if this is guaranteed to be always `1:1` for open files. @piotmag769?

mkaput: I've checked the source code and right now, it is `1:1`. The whole idea is that it'd be easier to change this in a single place than in all tests.

As for the comment, added `corresponding` word

Draggu (approved):

tests/e2e/support/mock_client.rs
piotmag769: >I'm not sure if this is guaranteed to be always 1:1 for open files.

@Draggu It is certainly not for Scarb projects, but for `cairo_project.toml` projects it is. We want to wait for all of them for `cairo_project.toml` projects since they mutate the db (even if they set the inputs to the same value they still bump the revision) causing cancellations.

Good catch, will remember about it while doing Scarb tests.

PR #44, opened 2024-12-18
Analyzing progress notifications

Arcticae:
Stack :
⬆️  #212 
⬇️  #190 

Based off this PR we can start implementing tests (like currently we can wait for project update + the diagnostics finish and it should work fine)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -101,6 +101,7 @@ impl Config {
                 debug!("reloaded configuration: {:#?}", state.config);
 
                 state.proc_macro_controller.on_config_change(&mut state.db, &state.config);
+                state.analysis_progress_controller.on_config_change(&state.config);
             })
         };
 
--- a/src/ide/analysis_progress.rs
+++ b/src/ide/analysis_progress.rs
@@ -1,0 +1,113 @@
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
+
+use crate::config::Config;
+use crate::server::client::Notifier;
+
+/// A struct that allows to track procmacro requests.
+#[derive(Clone)]
+pub struct ProcMacroRequestTracker {
+    procmacro_request_submitted: Arc<AtomicBool>,
+}
+
+impl ProcMacroRequestTracker {
+    pub fn new() -> Self {
+        Self { procmacro_request_submitted: Arc::new(AtomicBool::new(false)) }
+    }
+
+    /// Signals that a request to proc macro server was made during the current generation of
+    /// diagnostics.
+    pub fn register_procmacro_request(&self) {
+        self.procmacro_request_submitted.store(true, Ordering::SeqCst);
+    }
+
+    pub fn reset(&self) {
+        self.procmacro_request_submitted.store(false, Ordering::SeqCst);
+    }
+
+    pub fn get_did_submit_procmacro_request(&self) -> bool {
+        self.procmacro_request_submitted.load(Ordering::SeqCst)
+    }
+}
+
+#[derive(Clone)]
+pub struct AnalysisProgressController {
+    state: Arc<Mutex<AnalysisProgressControllerState>>,
+    request_tracker: ProcMacroRequestTracker,
+}
+
+impl AnalysisProgressController {
+    pub fn on_config_change(&self, config: &Config) {
+        self.state.lock().unwrap().on_config_change(config)
+    }
+
+    pub fn try_start_analysis(&self) {
+        self.request_tracker.reset();
+        self.state.lock().unwrap().try_start_analysis()
+    }
+
+    pub fn try_stop_analysis(&self) {
+        self.state
+            .lock()
+            .unwrap()
+            .try_stop_analysis(self.request_tracker.get_did_submit_procmacro_request());
+    }
+}
+
+impl AnalysisProgressController {
+    pub fn new(notifier: Notifier, request_tracker: ProcMacroRequestTracker) -> Self {
+        Self {
+            request_tracker,
+            state: Arc::new(Mutex::new(AnalysisProgressControllerState::new(notifier))),
+        }
+    }
+}
+
+/// Controller used to send notifications to the client about analysis progress.
+/// Uses information provided from other controllers (diagnostics controller, procmacro controller)
+/// to assess if diagnostics are in fact calculated.
+#[derive(Clone)]
+struct AnalysisProgressControllerState {
+    notifier: Notifier,
+    /// Indicates that a notification was sent and analysis (i.e. macro expansion) is taking place.
+    analysis_in_progress: bool,
+    /// Loaded asynchronously from config
+    procmacros_enabled: Option<bool>,
+}
+
+impl AnalysisProgressControllerState {
+    fn new(notifier: Notifier) -> Self {
+        Self { notifier, analysis_in_progress: false, procmacros_enabled: None }
+    }
+
+    pub fn on_config_change(&mut self, config: &Config) {
+        self.procmacros_enabled = Some(config.enable_proc_macros);
+    }
+
+    fn try_start_analysis(&mut self) {
+        if !self.analysis_in_progress {
+            self.analysis_in_progress = true;
+            #[cfg(feature = "testing")]
+            {
+                use crate::lsp::ext::testing::DiagnosticsCalculationStart;
+                self.notifier.notify::<DiagnosticsCalculationStart>(());
+            }
+        }
+    }
+
+    fn try_stop_analysis(&mut self, did_submit_procmacro_request: bool) {
+        let config_not_loaded = self.procmacros_enabled.is_none();
+        if (!did_submit_procmacro_request
+            || config_not_loaded
+            || (self.procmacros_enabled == Some(false)))
+            && self.analysis_in_progress
+        {
+            self.analysis_in_progress = false;
+            #[cfg(feature = "testing")]
+            {
+                use crate::lsp::ext::testing::DiagnosticsCalculationFinish;
+                self.notifier.notify::<DiagnosticsCalculationFinish>(());
+            }
+        }
+    }
+}
--- a/src/lang/proc_macros/client/mod.rs
+++ b/src/lang/proc_macros/client/mod.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, VecDeque};
+use std::fmt::{Debug, Formatter};
 use std::sync::{Mutex, MutexGuard};
 
 use anyhow::{Context, Result, anyhow, ensure};
@@ -16,6 +17,8 @@ use scarb_proc_macro_server_types::methods::expand::{
 pub use status::ClientStatus;
 use tracing::error;
 
+use crate::ide::analysis_progress::ProcMacroRequestTracker;
+
 pub mod connection;
 mod id_generator;
 pub mod status;
@@ -27,21 +30,26 @@ pub enum RequestParams {
     Inline(ExpandInlineMacroParams),
 }
 
-#[derive(Debug)]
 pub struct ProcMacroClient {
     connection: ProcMacroServerConnection,
     id_generator: id_generator::IdGenerator,
     requests_params: Mutex<HashMap<RequestId, RequestParams>>,
     error_channel: Sender<()>,
+    proc_macro_request_tracker: ProcMacroRequestTracker,
 }
 
 impl ProcMacroClient {
-    pub fn new(connection: ProcMacroServerConnection, error_channel: Sender<()>) -> Self {
+    pub fn new(
+        connection: ProcMacroServerConnection,
+        error_channel: Sender<()>,
+        proc_macro_request_tracker: ProcMacroRequestTracker,
+    ) -> Self {
         Self {
             connection,
             id_generator: Default::default(),
             requests_params: Default::default(),
             error_channel,
+            proc_macro_request_tracker,
         }
     }
 
@@ -146,6 +154,7 @@ impl ProcMacroClient {
 
         match self.send_request_untracked::<M>(id, &params) {
             Ok(()) => {
+                self.proc_macro_request_tracker.register_procmacro_request();
                 requests_params.insert(id, map(params));
             }
             Err(err) => {
@@ -159,6 +168,17 @@ impl ProcMacroClient {
     #[tracing::instrument(level = "trace", skip_all)]
     fn failed(&self) {
         let _ = self.error_channel.try_send(());
+    }
+}
+
+impl Debug for ProcMacroClient {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ProcMacroClient")
+            .field("connection", &self.connection)
+            .field("id_generator", &self.id_generator)
+            .field("requests_params", &self.requests_params)
+            .field("error_channel", &self.error_channel)
+            .finish_non_exhaustive()
     }
 }
 
--- a/src/lang/proc_macros/controller.rs
+++ b/src/lang/proc_macros/controller.rs
@@ -19,6 +19,7 @@ use super::client::connection::ProcMacroServerConnection;
 use super::client::status::ClientStatus;
 use super::client::{ProcMacroClient, RequestParams};
 use crate::config::Config;
+use crate::ide::analysis_progress::ProcMacroRequestTracker;
 use crate::lang::db::AnalysisDatabase;
 use crate::lang::proc_macros::db::ProcMacroGroup;
 use crate::lang::proc_macros::plugins::proc_macro_plugin_suite;
@@ -53,6 +54,7 @@ pub struct ProcMacroClientController {
     plugin_suite: Option<PluginSuite>,
     initialization_retries: RateLimiter<NotKeyed, InMemoryState, QuantaClock>,
     channels: ProcMacroChannels,
+    proc_macro_request_tracker: ProcMacroRequestTracker,
 }
 
 impl ProcMacroClientController {
@@ -60,10 +62,15 @@ impl ProcMacroClientController {
         self.channels.clone()
     }
 
-    pub fn new(scarb: ScarbToolchain, notifier: Notifier) -> Self {
+    pub fn new(
+        scarb: ScarbToolchain,
+        notifier: Notifier,
+        proc_macro_request_tracker: ProcMacroRequestTracker,
+    ) -> Self {
         Self {
             scarb,
             notifier,
+            proc_macro_request_tracker,
             plugin_suite: Default::default(),
             initialization_retries: RateLimiter::direct(
                 Quota::with_period(Duration::from_secs(
@@ -176,6 +183,7 @@ impl ProcMacroClientController {
                         self.channels.response_sender.clone(),
                     ),
                     self.channels.error_sender.clone(),
+                    self.proc_macro_request_tracker.clone(),
                 );
 
                 client.start_initialize();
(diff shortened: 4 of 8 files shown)


src/server/client.rs
```diff
@@ -26,7 +26,7 @@ pub struct Client<'s> {
     pub(super) requester: Requester<'s>,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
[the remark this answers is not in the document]
Arcticae: It's included in the new Controller which needs to be `Debug` as well because of `ProcMacroClient` being `Debug`

src/state.rs
```diff
@@ -48,23 +58,63 @@ impl State {
...
+        Beacon::wrap(SnapshotInternal {
             db: self.db.snapshot(),
             open_files: self.open_files.snapshot(),
             config: self.config.snapshot(),
+        })
+    }
+}
+
+pub struct Beacon<T> {
+    value: T,
```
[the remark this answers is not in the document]
Arcticae: Either i:
- do this
- add the method to the snapshot itself
- pass something around along with the snapshot (most changes in the code)

First 2 are mostly fine i think, 3rd one requires more intervention

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use crate::id_generator::IdGenerator;
+use crate::server::client::Notifier;
+use crate::state::Beacon;
+
+/// Controller used to send notifications to the client about analysis progress.
+/// Uses information provided from other controllers (diagnostics controller, procmacro controller)
+/// to assess if diagnostics are in fact calculated.
+#[derive(Debug, Clone)]
+pub struct AnalysisProgressController {
```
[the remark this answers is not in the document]
Arcticae: Can do although i think it's nicely explained here in the docstrings

src/server/client.rs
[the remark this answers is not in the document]
Draggu: Maybe custom impl on `ProcMacroClient` would be better in this case? This does not make sense to debug this struct

src/state.rs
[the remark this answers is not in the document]
Draggu: Discussed offline

src/state.rs
```diff
@@ -48,23 +58,63 @@ impl State {
...
-        StateSnapshot {
+        Beacon::wrap(SnapshotInternal {
             db: self.db.snapshot(),
             open_files: self.open_files.snapshot(),
             config: self.config.snapshot(),
+        })
+    }
+}
+
+pub struct Beacon<T> {
```
mkaput, line 77: is there a specific reason why are you wrapping `T` here?

can't this be a valueless object that is kept as a field in the old `StateSnapshot`? It would still be dropped whenever `StateSnapshot` is

mkaput (changes requested):

src/state.rs
Arcticae: 1. I resigned from the "on-drop" behaviour (because of lack of information when the analysis actually finished on a thread vs. the thread was cancelled) 
2. I can do this without wrapping - no problem there i think (discussion [here](https://github.com/software-mansion/cairols/pull/44#discussion_r1890257077))

Arcticae: Done

src/ide/analysis_progress.rs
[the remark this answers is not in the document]
Arcticae: TBD Still - i would like to finalize business logic first if that's OK so i don't have to update this flowchart

src/state.rs
```diff
@@ -37,8 +39,16 @@ impl State {
...
+
+        let analysis_progress_controller = AnalysisProgressController::new(notifier.clone());
+        let proc_macro_controller = ProcMacroClientController::new(
+            scarb_toolchain.clone(),
+            notifier.clone(),
+            analysis_progress_controller.clone(),
+        );
+
+        let diagnostics_controller =
+            DiagnosticsController::new(notifier.clone(), analysis_progress_controller.clone());
```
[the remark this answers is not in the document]
Arcticae: Done - pls let me know if that's what you meant

src/server/client.rs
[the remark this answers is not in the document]
Arcticae: Done

src/state.rs
```diff
@@ -68,6 +103,14 @@ pub struct StateSnapshot {
...
     pub open_files: Snapshot<HashSet<Url>>,
     pub config: Snapshot<Config>,
+    /// Beacon to signal when the snapshot is no longer used
+    pub beacon: Beacon,
+}
+
+impl StateSnapshot {
+    pub(crate) fn signal_finish(&mut self) {
+        self.beacon.signal();
+    }
```
piotmag769, line 113: Looking at the usage of `Beacon` is it worth introducing at all? You could do `pub callback`, set the callback (or private callback and have a setter) and then do `execute_callback` imo

src/state.rs
```diff
@@ -59,6 +71,29 @@ impl State {
...
+}
+/// Struct which allows setting a callback - which can be triggered afterward
+/// by the function which has the reference.
+#[derive(Default)]
+pub struct Beacon {
+    signal_hook: Option<Box<dyn FnOnce() + Send>>,
+}
+
+impl Beacon {
+    // Set the drop hook
```
piotmag769, line 86: dot.

Btw it doesn't set the drop hook, it is not done on `Drop`.

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,180 @@
...
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::{Arc, Mutex};
+
+use lsp_types::notification::Notification;
+
+use crate::id_generator::IdGenerator;
+use crate::server::client::Notifier;
+use crate::state::Beacon;
+
+/// A facade for `AnalysisProgressController` that allows to track progress of diagnostics
```
piotmag769, line 11: protostar_facade.jpg

src/state.rs
piotmag769: >I resigned from the "on-drop" behaviour (because of lack of information when the analysis actually finished on a thread vs. the thread was cancelled)

Can we add a comment about it somewhere near that?

src/state.rs
Arcticae: Discussed offline

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/// Notifies about diagnostics generation which ended calculating
+#[derive(Debug)]
+pub struct DiagnosticsCalculationFinish;
+
+impl Notification for DiagnosticsCalculationFinish {
+    type Params = ();
+    const METHOD: &'static str = "cairo/diagnosticsCalculationFinish";
+}
```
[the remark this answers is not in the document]
Arcticae: I don't know why not place it near the usage tbh

src/state.rs
```diff
@@ -59,6 +71,29 @@ impl State {
             db: self.db.snapshot(),
             open_files: self.open_files.snapshot(),
             config: self.config.snapshot(),
+            beacon: Default::default(),
+        }
+    }
+}
+/// Struct which allows setting a callback - which can be triggered afterward
+/// by the function which has the reference.
```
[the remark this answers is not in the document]
Arcticae: Reference to the thing that i'm talking about - the struct

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::state::Beacon;
+
+/// A facade for `AnalysisProgressController` that allows to track progress of diagnostics
+/// generation and procmacro requests.
+#[derive(Clone)]
+pub struct AnalysisProgressTracker {
+    controller: AnalysisProgressController,
+}
+
+impl AnalysisProgressTracker {
```
Draggu, line 18: We can split it for proc macros / diagnostics so interface is clear there, also it does not have to store whole controller (and should not if it possible, if not then maybe controller itself is ok).

src/state.rs
```diff
@@ -68,6 +103,14 @@ pub struct StateSnapshot {
     pub db: salsa::Snapshot<AnalysisDatabase>,
     pub open_files: Snapshot<HashSet<Url>>,
     pub config: Snapshot<Config>,
+    /// Beacon to signal when the snapshot is no longer used
+    pub beacon: Beacon,
+}
+
+impl StateSnapshot {
+    pub(crate) fn signal_finish(&mut self) {
```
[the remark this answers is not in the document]
Draggu: Discussed offline

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,180 @@
...
+            generation_id: Arc::new(AtomicU64::new(id_generator.unique_id())),
+        }
+    }
+
+    pub fn set_did_submit_procmacro_request(&self, value: bool) {
+        self.did_submit_procmacro_request.store(value, Ordering::SeqCst);
+    }
+
+    /// Allows to set the procmacro configuration to whatever is in the config, upon loading it.
+    pub fn set_procmacros_enabled(&self, value: bool) {
```
Draggu, line 92:
```suggestion
    pub fn on_config_change(&self, config: &Config) {
```

More future proof and it is clear where this info is coming from.

src/lang/proc_macros/client/mod.rs
```diff
@@ -142,6 +150,7 @@ impl ProcMacroClient {
         match self.send_request_untracked::<M>(id, &params) {
             Ok(()) => {
                 requests_params.insert(id, map(params));
+                self.analysis_progress_tracker.register_procmacro_request();
```
piotmag769, line 153: I believe that there is a small chance that a yet alive snapshot from previous diagnostics calculation was already in `db.get_attribute_expansion` before its cancellation. It will still call this method and possibly register proc macro request in analysis tracker AFTER `track_analysis` is called to start tracking the new diagnostics batch. It can cause us to never finish waiting for diagnostics in tests, since:
- the next batch won't be scheduled
- we will always think there is some proc macro yet to resolve while it may not be

piotmag769 (changes requested):

src/state.rs
piotmag769: Bcs I was a second person that asked to implement it on drop xD

src/state.rs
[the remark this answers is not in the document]
piotmag769: Can we state that explicitly in the comment? It can be figured out, but it is not immediately obvious on the (English) language level

src/ide/analysis_progress.rs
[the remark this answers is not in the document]
piotmag769: Both are fine for me, but we mostly go with the approach I listed. The only time ur approach is used is for scarb resolving notifications.

src/ide/analysis_progress.rs
piotmag769, line 1: A lot of pubs are unnecessary here, some methods could be inlined

piotmag769 (commented):
Logic wise seems good besides what I wrote before

src/state.rs
```diff
@@ -59,6 +71,29 @@ impl State {
             db: self.db.snapshot(),
             open_files: self.open_files.snapshot(),
             config: self.config.snapshot(),
+            beacon: Default::default(),
+        }
+    }
+}
+/// Struct which allows setting a callback - which can be triggered afterward
```
mkaput, line 78:
```suggestion
}

/// Struct which allows setting a callback - which can be triggered afterward
```

src/state.rs
```diff
@@ -59,6 +71,29 @@ impl State {
...
+/// by the function which has the reference.
+#[derive(Default)]
+pub struct Beacon {
```
mkaput, line 81: TBH you have simplified this so much that I would probably drop this type entirely and just stick that `Option<Box<dyn FnOnce() + Send>>`.

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// diagnostics.
+    pub fn register_procmacro_request(&self) {
+        self.controller.set_did_submit_procmacro_request(true);
+    }
+
+    /// Sets handlers for tracking beacons sent to threads.
+    /// The beacons are wrapping snapshots, which are signalling when diagnostics finished
+    /// calculating for a given snapshot (used for calculating files diagnostics or removing
+    /// stale ones)
+    pub fn track_analysis<'a>(&self, beacons: impl Iterator<Item = &'a mut Beacon>) {
```
mkaput, line 29: Can't this just take a mutable reference to StateSnapshots?

mkaput (changes requested):

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,180 @@
...
+pub struct AnalysisProgressController {
+    notifier: Notifier,
+    /// ID of the diagnostics "generation" - the scheduled diagnostics jobs set.
+    /// Used to filter out stale threads finishing when new ones (from newer "generation")
+    /// are already in progress and being tracked by the controller.
+    generation_id: Arc<AtomicU64>,
+    /// Sequential IDs of state snapshots from the current generation, used to track their status
+    /// (present meaning it's still being used)
+    active_snapshots: Arc<Mutex<HashSet<usize>>>,
+    id_generator: Arc<IdGenerator>,
```
mkaput, line 58: TBH you'd be better off not touching the IdGenerator here and just peeing that AtomicU64 in generation_id and incrementing it directly. you're just doing this in a fancy way now

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// If `true` - a request to procmacro server was submitted, meaning that analysis will extend
+    /// beyond the current generation of diagnostics.
+    did_submit_procmacro_request: Arc<AtomicBool>,
+    /// Indicates that a notification was sent and analysis (i.e. macro expansion) is taking place.
+    analysis_in_progress: Arc<AtomicBool>,
+    /// Loaded asynchronously from config - unset if config was not loaded yet.
+    /// Has to be set in order for analysis to finish.
+    procmacros_enabled: Arc<Mutex<Option<bool>>>,
+}
```
mkaput, line 67: consider just using a mutex for everything. I don't believe you benefit from atomics here over mutexes, and I'm not sure all methods here guarantee consistency in case of races.

```rust
struct AnalysisProgressController(Arc<Mutex<AnalysisProgressControllerState>>);

struct AnalysisProgressControllerState {
    generation_id: u64,
    // ...
}
```

src/ide/analysis_progress.rs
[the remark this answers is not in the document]
mkaput: let's place it in `lsp/ext.rs` as discussed on meeting

mkaput (changes requested):

src/lang/proc_macros/controller.rs
```diff
@@ -176,10 +183,10 @@ impl ProcMacroClientController {
                         self.channels.response_sender.clone(),
                     ),
                     self.channels.error_sender.clone(),
+                    self.proc_macro_request_tracker.clone(),
                 );
 
                 client.start_initialize();
-
```
Draggu, line 182: new line

src/config.rs
```diff
@@ -97,10 +97,10 @@ impl Config {
                     response.pop_front().as_ref().and_then(Value::as_bool).unwrap_or_default();
                 state.config.enable_proc_macros =
                     response.pop_front().as_ref().and_then(Value::as_bool).unwrap_or(false);
-
```
Draggu, line 100: new line

Draggu (approved):

src/lsp/ext.rs
```diff
@@ -89,6 +89,24 @@ pub mod testing {
...
+}
+
+/// Notifies about diagnostics generation which ended calculating
+#[derive(Debug)]
+pub struct DiagnosticsCalculationFinish;
+
+impl Notification for DiagnosticsCalculationFinish {
+    type Params = ();
+    const METHOD: &'static str = "cairo/diagnosticsCalculationFinish";
+}
```
mkaput, line 108: we wanted to have a single "analysis status" notificaiton, didn't we?

src/lang/proc_macros/client/mod.rs
```diff
@@ -162,6 +171,17 @@ impl ProcMacroClient {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ProcMacroClient")
+            .field("connection", &self.connection)
+            .field("id_generator", &self.id_generator)
+            .field("requests_params", &self.requests_params)
+            .field("error_channel", &self.error_channel)
+            .finish_non_exhaustive()
+    }
+}
+
```
mkaput, line 184: hmm, what about adding a dummy Debug impl to ProcMacroRequestTracker that prints just `ProcMacroRequestTracker { .. }`, wouldn't it make things simpler?

mkaput (changes requested):

src/lsp/ext.rs
Arcticae: It's the concept in RA really 
Do you believe that it will be more future proof?

src/lang/proc_macros/client/mod.rs
Arcticae: Why not debug this as well though?

src/lsp/ext.rs
mkaput: I think that's what we wanted, to have an enum of states like `Idle`, `ExpandingProcMacros`, `Analysing`. This would be troublesome to implement if we start doing `x started`/`x finished` notifications.

src/lang/proc_macros/client/mod.rs
mkaput: I recall we discussed we do not want to debug its internals anyway. But anyhow, this is subjective idea, I'm not even attached to it myself, ignore this comment if you wish.

mkaput (approved):

piotmag769 (approved):

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,113 @@
...
+/// Controller used to send notifications to the client about analysis progress.
+/// Uses information provided from other controllers (diagnostics controller, procmacro controller)
+/// to assess if diagnostics are in fact calculated.
+#[derive(Clone)]
+pub struct AnalysisProgressControllerState {
+    notifier: Notifier,
+    /// Indicates that a notification was sent and analysis (i.e. macro expansion) is taking place.
+    analysis_in_progress: bool,
+    /// Loaded asynchronously from config
+    procmacros_enabled: Option<bool>,
```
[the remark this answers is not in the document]
Arcticae: It's more correct this way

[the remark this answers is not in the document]
piotmag769: Idk how tbh, you just store an extra info that you don't need. In config we also just have `bool`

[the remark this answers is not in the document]
Arcticae: > Idk how tbh

It's signaling that config hasn't been loaded yet. If you wanted to have specific behavior here that relies on the fact that config hasn't been loaded (i.e. not track any analysis before that) you won't be able to do that. 

> In config we also just have bool

If i could change that i would - but here it's out of scope

src/ide/analysis_progress.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl AnalysisProgressControllerState {
+    fn new(notifier: Notifier) -> Self {
+        Self { notifier, analysis_in_progress: false, procmacros_enabled: None }
+    }
+
+    pub fn on_config_change(&mut self, config: &Config) {
```
piotmag769, line 83: You don't need the pub too

PR #133, opened 2025-01-02
Add method to get LS and Scarb version

Draggu:
In response for https://github.com/software-mansion/vscode-cairo/pull/34#discussion_r1891217530

### DIFF
--- a/src/lsp/ext.rs
+++ b/src/lsp/ext.rs
@@ -1,5 +1,7 @@
 //! CairoLS extensions to the Language Server Protocol.
 
+use std::path::PathBuf;
+
 use lsp_types::notification::Notification;
 use lsp_types::request::Request;
 use lsp_types::{TextDocumentPositionParams, Url};
@@ -49,6 +51,28 @@ pub struct CorelibVersionMismatch;
 impl Notification for CorelibVersionMismatch {
     type Params = String;
     const METHOD: &'static str = "cairo/corelib-version-mismatch";
+}
+
+/// Collects versions of LS and it's dependencies.
+#[derive(Debug)]
+pub struct ToolchainInfo;
+
+#[derive(Serialize, Deserialize)]
+pub struct PathAndVersion {
+    pub path: PathBuf,
+    pub version: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct ToolchainInfoResponse {
+    pub ls: PathAndVersion,
+    pub scarb: Option<PathAndVersion>,
+}
+
+impl Request for ToolchainInfo {
+    type Params = ();
+    type Result = ToolchainInfoResponse;
+    const METHOD: &'static str = "cairo/toolchainInfo";
 }
 
 #[cfg(feature = "testing")]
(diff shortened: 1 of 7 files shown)


src/server/routing/mod.rs
```diff
@@ -44,6 +44,9 @@ pub fn request<'a>(request: Request) -> Task<'a> {
         ExpandMacro::METHOD => {
             background_request_task::<ExpandMacro>(request, BackgroundSchedule::Worker)
         }
+        LSVersions::METHOD => {
```
mkaput, line 47: nitpick: can you please put this in alphabetical order?

src/lsp/ext.rs
```diff
@@ -51,6 +53,28 @@ impl Notification for CorelibVersionMismatch {
...
 
+/// Collects versions of LS and it's dependencies.
+#[derive(Debug)]
+pub struct LSVersions;
+
+#[derive(Serialize, Deserialize)]
+pub struct BinWithVersion {
+    pub version: String,
+    pub path: PathBuf,
+}
```
mkaput, line 64:
```suggestion
#[derive(Serialize, Deserialize)]
pub struct PathAndVersion {
    pub path: PathBuf,
    pub version: String,
}
```

wdyt?

1. Name follows definition, which actually makes sense given this is just a typed named tuple.
2. I feel `with` is not something an English-native would say here.

src/lsp/ext.rs
```diff
@@ -51,6 +53,28 @@ impl Notification for CorelibVersionMismatch {
     const METHOD: &'static str = "cairo/corelib-version-mismatch";
 }
 
+/// Collects versions of LS and it's dependencies.
+#[derive(Debug)]
+pub struct LSVersions;
```
mkaput, line 58:
```suggestion
pub struct ToolchainInfo;
```

to me it sounds like we might want to stick more information in this struct in the future and this name looks more general to me. wdyt?

src/project/mod.rs
```diff
@@ -22,6 +23,7 @@ use crate::state::State;
 use crate::toolchain::scarb::ScarbToolchain;
 
 mod crate_data;
+mod ls_versions;
```
mkaput, line 26: I feel like this would be better placed as `crate::toolchain::info` module.

src/state.rs
```diff
@@ -66,6 +67,7 @@ impl State {
 /// Readonly snapshot of Language server state.
 pub struct StateSnapshot {
     pub db: salsa::Snapshot<AnalysisDatabase>,
+    pub scarb_toolchain: Snapshot<ScarbToolchain>,
```
mkaput, line 70:
```suggestion
    pub scarb_toolchain: ScarbToolchain,
```

`ScarbToolchain` is a sync singleton. We don't need `Owned`/`Snapshot` system for it imho

src/toolchain/scarb.rs
```diff
@@ -166,6 +175,22 @@ impl ScarbToolchain {
 
         Ok(proc_macro_server)
     }
+
+    pub fn version(&self) -> Option<String> {
+        self.version.get_or_init(|| self.fetch_version().ok()).clone()
```
mkaput, line 180: `error!` the `Result` here

mkaput (changes requested):

mkaput (approved):

piotmag769 (approved):

PR #167, opened 2025-01-13
Scarb test: removing workspace member

piotmag769:
**Stack**:
- #175
- #169
- #168
- #167 ⬅

### DIFF
--- a/tests/e2e/scarb/scarb_toml_change.rs
+++ b/tests/e2e/scarb/scarb_toml_change.rs
@@ -1,0 +1,78 @@
+use cairo_lang_test_utils::parse_test_file::TestRunnerResult;
+use cairo_lang_utils::ordered_hash_map::OrderedHashMap;
+use cairo_language_server::lsp;
+use lsp_types::notification::DidChangeWatchedFiles;
+use lsp_types::{
+    ClientCapabilities, DidChangeWatchedFilesClientCapabilities, DidChangeWatchedFilesParams,
+    FileChangeType, FileEvent, WorkspaceClientCapabilities,
+};
+use similar::TextDiff;
+
+use crate::support::normalize::normalize;
+use crate::support::sandbox;
+
+cairo_lang_test_utils::test_file_test!(
+    removing_member,
+    "tests/test_data/scarb",
+    {
+        removing_member: "removing_member.txt"
+    },
+    test_removing_member
+);
+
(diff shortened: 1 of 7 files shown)


tests/e2e/scarb/scarb_toml_change.rs
```diff
@@ -0,0 +1,79 @@
...
+    assert!(ls.wait_for_diagnostics("b/src/lib.cairo").diagnostics.is_empty());
+
+    let analyzed_crates = ls.send_request::<lsp::ext::ViewAnalyzedCrates>(());
+
+    ls.edit_file("Scarb.toml", indoc! {r#"
+        [workspace]
+        members = [
+            "a",
+        ]
+    "#
```
mkaput, line 63: consider moving this as an input, say: `Scarb.toml (modified)`

mkaput (approved):

tests/test_data/scarb/removing_member.txt
```diff
@@ -0,0 +1,259 @@
...
+        },
+        experimental_features: ExperimentalFeaturesConfig {
+            negative_impls: true,
+            associated_item_constraints: true,
+            coupons: true,
+        },
+    }
+    ```
+
+//! > Analyzed crates after member removal
```
mkaput, line 145: a non-obligatory idea, wdyt: instead of slapping changed output as-is here, what if we save a diff?

tests/e2e/support/mock_client.rs
```diff
@@ -318,6 +318,10 @@ impl MockClient {
         );
     }
 
+    pub fn edit_file(&mut self, path: impl AsRef<Path>, contents: &str) {
```
Draggu, line 321: This is modifying the file without notifying LS about it, please explain why it is done like this. In a real life scenario, I believe it would trigger `workspace/didChangeWatchedFiles`

tests/test_data/scarb/removing_member.txt
piotmag769: Done

tests/e2e/support/mock_client.rs
piotmag769: It is triggered manually in tests, it is on the caller to guarantee sending a correct notification

Draggu (approved):

PR #176, opened 2025-01-14
Reject text matches that are not at word boundaries

mkaput:
**Stack**:
- #192
- #188
- #186
- #185
- #178
- #177
- #176 ⬅

### DIFF
--- a/src/lang/inspect/usages.rs
+++ b/src/lang/inspect/usages.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_filesystem::db::FilesGroup;
 use cairo_lang_filesystem::ids::FileId;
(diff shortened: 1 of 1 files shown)


src/lang/inspect/usages.rs
```diff
@@ -98,6 +96,28 @@ impl<'a> FindUsages<'a> {
         files.into_iter()
     }
 
+    fn match_offsets<'b>(
```
piotmag769, line 99: [nit]
```suggestion
    fn offsets_of_text_matches<'b>(
```
or `text_matches_offsets`. `match_offsets` suggests it matches some offsets

piotmag769 (approved):

src/lang/inspect/usages.rs
mkaput: I prefer my form, because `match` is a verb here that states what action is done

integraledelebesgue (approved):

PR #180, opened 2025-01-14
Rework expand macro tests with `insta.rs`

Draggu:


tests/e2e/support/load_cairo.rs
```diff
@@ -0,0 +1,22 @@
+use std::path::PathBuf;
+
+#[macro_export]
+macro_rules! load_cairo {
+    ($file:expr) => {
+        $crate::support::load_cairo::load_cairo(file!(), $file)
+    };
+}
+
+pub fn load_cairo(current_file: &str, file_to_include: &str) -> String {
```
piotmag769, line 10: 👉 👈 uwu
```suggestion
pub fn load_cairo(current_file: impl AsRef<Path>, file_to_include: impl AsRef<Path>) -> String {
```

piotmag769 (approved):

mkaput (changes requested):

tests/e2e/support/load_cairo.rs
Draggu: `current_file` is always `&str` so I would rather stick with the current form, `file_to_include` is transformed into owned `PathBuf` so `impl Into<PathBuf>` is better.

integraledelebesgue (approved):

tests/e2e/support/load_data.rs
```diff
@@ -0,0 +1,21 @@
...
+
+macro_rules! data {
+    ($file:expr) => {
+        $crate::support::load_data::load_data(file!(), $file)
+    };
+}
+
+pub(crate) use data;
+
+pub fn load_data(current_file: &str, file_to_include: impl Into<PathBuf>) -> String {
```
mkaput, line 11:
```suggestion
#[doc(hidden)]
pub fn data(current_file: &str, file_to_include: impl Into<PathBuf>) -> String {
```

tests/e2e/support/mod.rs
```diff
@@ -3,6 +3,7 @@ pub mod client_capabilities;
 pub mod cursor;
 pub mod fixture;
 pub mod jsonrpc;
+pub mod load_data;
```
mkaput, line 6: either

```suggestion
pub mod data;
```

or maybe

```suggestion
pub mod insta;
```

?

tests/e2e/macro_expand/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+        .chunk_by(|report| report.1.clone())
+        .into_iter()
+        .enumerate()
+    {
+        let source_contexts =
+            source_contexts.map(|(source_context, _)| source_context).collect_vec().join("\n");
+
+        let report = MacroExpandReport { source_contexts, expansion };
+
+        insta::assert_toml_snapshot!(format!("{case_name}#{n}"), report);
```
mkaput, line 68: after working on #188 I have one, I think important, observation:

calling insta snapshots multiple times in a test looks like an antipattern. it is very disruptive to the `cargo test -> cargo insta review` cycle because first non-matching snapshot kills entire test and prevents later snapshots to actually test anything and this quickly becomes annoying

think about avoiding doing loops here

mkaput (changes requested):

tests/e2e/support/load_data.rs
Draggu: This will lead to ident collision with macro because of
```
pub(crate) use data;
```

mkaput: ok, but please add doc(hidden)

tests/e2e/macro_expand/mod.rs
Draggu: Discussed offline

tests/e2e/support/load_data.rs
mkaput: not done

Draggu: Was sure that added this lol

mkaput (approved):

Arcticae (approved):

PR #190, opened 2025-01-16
Synchronized diagnostics worker threads

Arcticae:
Stack ⬆️ : #44  

In order to achieve more robust tracking of diagnostics progress, we need to make those threads converge. 

That means two different generations' threads now don't run in parallel, and modify state (in turn making the progress tracking very difficult).


src/lang/diagnostics/mod.rs
```diff
@@ -109,29 +114,41 @@ impl DiagnosticsControllerThread {
...
+        worker_status_receiver.iter().nth(jobs_number - 1);
     }
 
     /// Shortcut for spawning a worker task which does the boilerplate around cloning state parts
     /// and catching panics.
-    fn spawn_worker(&self, f: impl FnOnce(ProjectDiagnostics, Notifier) + Send + 'static) {
+    fn spawn_worker(
+        &self,
+        f: impl FnOnce(ProjectDiagnostics, Notifier) + Send + 'static,
+        sender: crossbeam::channel::Sender<WorkerStatus>,
```
mkaput, line 150:
```suggestion
        sender: crossbeam::channel::Sender<()>,
```

TBH I don't see a need to declare a custom ZST for this. I would rather focus on encapsulating entire waiting logic into some kind of `TaskHandle` with a `join` method, like we have for threads

src/lang/diagnostics/mod.rs
```diff
@@ -143,23 +160,32 @@ impl DiagnosticsControllerThread {
                     error!("caught panic in diagnostics worker");
                 }
             }
+            sender.send(WorkerStatus::Stopped).unwrap();
```
mkaput, line 163: I'm afraid of two facts here:
1. This is panicking
2. This is not called from a drop guard, meaning it is possible that some cautionless code change may cause this to never be called.

mkaput (changes requested):

src/lang/diagnostics/mod.rs
Arcticae: 1. Yes but it shouldn't unless we do something wrong 
2. It will be clear when it happens when we have any diagnostics tests since diagnostics will not be reported

mkaput: Ad 1. Please replace this unwrap with an `expect` stating this should never happen "because"
Ad 2. Not sure I understand, can you rephrase plz?

Arcticae: This is no longer relevant after rewrite of logic to `TaskHandle`

src/lang/diagnostics/mod.rs
Arcticae: Done

piotmag769: I believe @Draggu did it this way in the first diagnostic impls. It was for better reasoning about what to send if sth was cancelled etc. (e.g. we can still send sth if cancellation happened if diags for some files were calculated before cancellation etc.)

src/lang/diagnostics/mod.rs
```diff
@@ -143,15 +151,20 @@ impl DiagnosticsControllerThread {
                     error!("caught panic in diagnostics worker");
                 }
             }
-        });
+        })
     }
 
     /// Makes batches out of `files` and spawns workers to run [`refresh_diagnostics`] on them.
-    fn spawn_refresh_worker(&self, files: &[FileId], state_snapshots: Vec<StateSnapshot>) {
+    fn spawn_refresh_worker(
```
piotmag769, line 158:
```suggestion
    fn spawn_refresh_workers(
```

piotmag769 (approved):

src/server/schedule/thread/pool.rs
```diff
@@ -48,6 +49,37 @@ struct Job {
     f: Box<dyn FnOnce() + Send + 'static>,
 }
 
+struct TaskState {
+    finished: bool,
+}
+
+#[derive(Clone)]
+pub struct TaskHandle {
+    state: Arc<(Mutex<TaskState>, Condvar)>,
```
Draggu, line 58: You can use `trigger` here

Draggu (changes requested):

src/lang/diagnostics/mod.rs
```diff
@@ -109,29 +109,37 @@ impl DiagnosticsControllerThread {
...
     fn diagnostics_controller_tick(&self, state_snapshots: StateSnapshots) {
+        let mut worker_handles = Vec::new();
         let (state, primary_snapshots, secondary_snapshots) = state_snapshots.split();
 
         let primary_set = find_primary_files(&state.db, &state.open_files);
         let primary: Vec<_> = primary_set.iter().copied().collect();
-        self.spawn_refresh_worker(&primary, primary_snapshots);
+        worker_handles.extend(self.spawn_refresh_worker(&primary, primary_snapshots));
 
         let secondary = find_secondary_files(&state.db, &primary_set);
```
piotmag769, line 119: Non-working case:

There is a chance that the a cancellation happens here, we catch it in the event loop, we log the cancel and then move to the next while iteration. But we already spawned workers for primary files and we won't wait for them.

We should instead store `workers_handles` in `DiagnosticsControllerThread` and wait for the threads to finish in the event loop, then clear the `workers_handles`.

piotmag769 (changes requested):

src/lang/diagnostics/mod.rs
Draggu: In first impl we were sending update and then applying it to state that lived on diagnostics main thread, now we are using `RwLock` for this so it does not matter

src/server/schedule/thread/pool.rs
```diff
@@ -48,6 +51,28 @@ struct Job {
     f: Box<dyn FnOnce() + Send + 'static>,
 }
 
+#[derive(Clone)]
+pub struct TaskHandle {
+    state: Arc<(trigger::Sender<()>, trigger::Receiver<()>)>,
```
Draggu, line 56: `Sender` and `Receiver` are wrappers on `Arc` so no need for another one, and as so u can use 2 props for this

```
sender: Sender<()>,
receiver: Receiver<()>,
```

src/lang/diagnostics/mod.rs
Arcticae: Done

src/lang/diagnostics/mod.rs
```diff
@@ -103,20 +106,25 @@ impl DiagnosticsControllerThread {
                     error!("caught panic while refreshing diagnostics");
                 }
             }
+
+            for handle in self.worker_handles.iter() {
```
Draggu, line 110: Calling `.join()` on all handlers and then clearing list could be moved to separate method like `wait_for_all` or something, up to you

src/server/schedule/thread/pool.rs
Arcticae: Done

src/lang/diagnostics/mod.rs
Arcticae: Done - thx for pointing this out 🙏

src/server/schedule/thread/pool.rs
```diff
@@ -94,19 +119,23 @@ impl Pool {
...
     {
+        let handle = TaskHandle::new();
+        let handle_clone = handle.clone();
         let f = Box::new(move || {
             if cfg!(debug_assertions) {
                 priority.assert_is_used_on_current_thread();
             }
             f();
+            handle_clone.signal_finish();
         });
```
Draggu, line 134:
```suggestion
        let handle = TaskHandle::new();
        let f = Box::new({
            let handle = handle.clone();
        
            move || {
                if cfg!(debug_assertions) {
                    priority.assert_is_used_on_current_thread();
                }
                f();
                handle_clone.signal_finish();
            }
        });
```

What do you think? Feel free to resolve.

src/lang/diagnostics/mod.rs
Arcticae: @piotmag769 @mkaput can we resolve?

src/server/schedule/thread/pool.rs
Arcticae: It's more clear to me as it was tbh

src/server/schedule/thread/pool.rs
Arcticae: done

src/lang/diagnostics/mod.rs
Arcticae: done

src/lang/diagnostics/trigger.rs
```diff
@@ -10,11 +10,13 @@ const POISON_PANIC: &str = "invariant error: trigger mutex should never become p
 /// The writer side of a trigger sender-receiver pair.
 ///
 /// See [`trigger`] for more information.
+#[derive(Clone)]
 pub struct Sender<T>(Arc<Inner<T>>);
 
 /// The reader side of a trigger sender-receiver pair.
 ///
 /// See [`trigger`] for more information.
+#[derive(Clone)]
```
Draggu, line 19: This can not be `Clone`

>        // Wake up a thread blocked by waiting on the corresponding receiver.
>        condvar.notify_one();

Arcticae: so it's not possible to keep this separately in the state like you commented before

Draggu: Split `TaskHandle` into 2 structs, then there will be no need to clone receiver, sender can be cloned

src/lang/diagnostics/trigger.rs
```diff
@@ -9,12 +9,15 @@ const POISON_PANIC: &str = "invariant error: trigger mutex should never become p
 
 /// The writer side of a trigger sender-receiver pair.
 ///
-/// See [`trigger`] for more information.
+/// See [`trigger`] for more information
+/// NOTE: Do not clone this or #[derive(Clone)], as this might result in unexpected behavior on
```
Draggu, line 13: We can have multiple senders, We can't have multiple receivers as we are using `notify_one`

src/lang/diagnostics/mod.rs
mkaput: yes, we can

src/lang/diagnostics/mod.rs
Arcticae: What if i sometime want to pass a value on finish? 
Plus, i don't like solutions with implicit behavior like "on drop" (which is ultimately about resource freeing and not hooking business logic onto that)

src/server/schedule/thread/pool.rs
```diff
@@ -28,6 +28,8 @@ use std::thread::available_parallelism;
 use crossbeam::channel;
 
 use super::{Builder, JoinHandle, ThreadPriority};
+use crate::lang::diagnostics::trigger;
+use crate::server::schedule::thread::pool::trigger::trigger;
```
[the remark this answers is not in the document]
Arcticae: done

src/server/schedule/thread/pool.rs
```diff
@@ -48,6 +50,30 @@ struct Job {
     f: Box<dyn FnOnce() + Send + 'static>,
 }
 
+#[derive(Clone)]
+pub struct TaskHandle(trigger::Receiver<()>);
+pub struct TaskTracker(trigger::Sender<()>);
+
+impl TaskTracker {
+    /// Signals that a task finished executing.
+    pub fn signal_finish(&self) {
```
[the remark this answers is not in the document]
Arcticae: please refer to my comment in the thread above why i'm reluctant to do this

src/server/schedule/thread/pool.rs
```diff
@@ -48,6 +50,30 @@ struct Job {
...
+        self.0.wait();
+    }
+}
+
+/// Creates single message channel for making it possible to wait for finishing tasks execution.
+pub fn task_progress_monitor() -> (TaskTracker, TaskHandle) {
+    let (sender, receiver) = trigger::<()>();
+    (TaskTracker(sender), TaskHandle(receiver))
+}
+
```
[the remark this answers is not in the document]
Arcticae: done

src/lang/diagnostics/trigger.rs
Arcticae: not relevant anymore

src/lang/diagnostics/trigger.rs
Arcticae: done

piotmag769 (approved):

src/server/trigger.rs
```diff
@@ -9,7 +9,7 @@ const POISON_PANIC: &str = "invariant error: trigger mutex should never become p
 
 /// The writer side of a trigger sender-receiver pair.
 ///
-/// See [`trigger`] for more information.
```
Draggu, line 12: dot

Draggu (approved):

mkaput (approved):

PR #197, opened 2025-01-17
Fix definition node for functions

mkaput:
**Stack**:
- #222
- #197 ⬅

### DIFF
--- a/src/lang/defs.rs
+++ b/src/lang/defs.rs
@@ -2,8 +2,8 @@ use std::iter;
 
 use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_defs::ids::{
-    FunctionTitleId, LanguageElementId, LookupItemId, MemberId, ModuleId, ModuleItemId,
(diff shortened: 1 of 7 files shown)


piotmag769 (approved):

src/lang/defs.rs
```diff
@@ -686,33 +687,63 @@ fn resolved_generic_item_def(
...
+                GenericFunctionId::Trait(id) => id
+                    .trait_function(db.upcast())
+                    .stable_ptr(db.upcast())
+                    .lookup(db.upcast())
+                    .declaration(db.upcast()),
             };
-            title.untyped_stable_ptr(db.upcast())
+            declaration.name(db.upcast()).stable_ptr().untyped()
         }
+
```
Draggu, line 720: Why all of these newlines?

mkaput: I am successively expanding each branch in this `match` expr and it starts being unpleasant to look at. Very subjective topic, can revert if you think I should do.

Draggu (approved):

PR #209, opened 2025-01-21
Store plugins in the Crate structure

integraledelebesgue:
**Stack**:
- #210
- #209 ⬅

### DIFF
--- a/src/project/crate_data.rs
+++ b/src/project/crate_data.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::path::PathBuf;
 
 use cairo_lang_defs::db::DefsGroup;
@@ -10,6 +11,7 @@ use cairo_lang_filesystem::ids::{CrateId, CrateLongId, Directory};
 use cairo_lang_utils::{Intern, LookupIntern};
 use smol_str::SmolStr;
 
+use super::builtin_plugins::BuiltinPlugin;
 use crate::lang::db::AnalysisDatabase;
 
 /// A complete set of information needed to set up a real crate in the analysis database.
@@ -36,6 +38,10 @@ pub struct Crate {
 
     /// Crate settings.
     pub settings: CrateSettings,
+
+    /// Built-in plugins required by the crate.
+    #[allow(dead_code)] // TODO: remove. The field is used later in the stack
+    pub builtin_plugins: HashSet<BuiltinPlugin>,
 }
 
 impl Crate {
@@ -61,6 +67,9 @@ impl Crate {
         if let Some(file_stems) = &self.custom_main_file_stems {
             inject_virtual_wrapper_lib(db, crate_id, file_stems);
         }
+
+        // TODO (later in the stack): Construct the plugin suite
+        // from builtin_plugins and set as an override for the crate.
     }
 
     /// Construct a [`Crate`] from data already applied to the [`AnalysisDatabase`].
@@ -79,7 +88,11 @@ impl Crate {
 
         let custom_main_file_stems = extract_custom_file_stems(db, crate_id);
 
-        Some(Self { name, discriminator, root, custom_main_file_stems, settings })
+        // TODO (later in the stack): Extract plugins associated with this crate
+        // from db and store it in this set.
+        let builtin_plugins = HashSet::default();
+
+        Some(Self { name, discriminator, root, custom_main_file_stems, settings, builtin_plugins })
     }
 
     /// States whether this is the `core` crate.
--- a/src/project/scarb.rs
+++ b/src/project/scarb.rs
@@ -9,12 +9,13 @@ use cairo_lang_filesystem::db::{
 };
 use itertools::Itertools;
 use scarb_metadata::{
-    CompilationUnitComponentDependencyMetadata, CompilationUnitComponentId, Metadata,
-    PackageMetadata,
+    CompilationUnitCairoPluginMetadata, CompilationUnitComponentDependencyMetadata,
+    CompilationUnitComponentId, Metadata, PackageMetadata,
 };
 use smol_str::ToSmolStr;
 use tracing::{debug, error, warn};
 
+use super::builtin_plugins::BuiltinPlugin;
 use super::manifest_registry::member_config::MemberConfig;
 use crate::lang::db::AnalysisDatabase;
 use crate::project::crate_data::Crate;
@@ -183,12 +184,53 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
 
             let custom_main_file_stems = (file_stem != "lib").then_some(vec![file_stem.into()]);
 
+            let plugin_dependencies = component
+                .dependencies
+                .as_ref()
+                .map(|dependencies| {
+                    dependencies
+                        .iter()
+                        .filter_map(|dependency| {
+                            let plugin = compilation_unit.cairo_plugins.iter().find(|plugin| {
+                                plugin.component_dependency_id.as_ref() == Some(&dependency.id)
+                            });
+
+                            if plugin.is_none() {
+                                error!(
+                                    "Plugin dependency `{}` not found in Scarb metadata",
+                                    dependency.id
+                                );
+                            }
+
+                            plugin
+                        })
+                        .collect_vec()
+                })
+                .unwrap_or_default();
+
+            // We collect only the built-in plugins.
+            // Procedural macros are handled separately in the
+            // `crate::lang::proc_macros::controller`.
+
+            let mut builtin_plugins = crates_by_component_id
+                .get(&component_id)
+                .map(|cr| cr.builtin_plugins.clone())
+                .unwrap_or_default();
+
+            builtin_plugins.extend(if package.is_some_and(|p| p.name == "core") {
+                // Corelib is a special case because it is described by `cairo_project.toml`.
+                plugins_for_corelib()
+            } else {
+                plugins_from_dependencies(metadata, &plugin_dependencies)
+            });
+
             let cr = Crate {
                 name: crate_name.into(),
                 discriminator: component.discriminator.as_ref().map(ToSmolStr::to_smolstr),
                 root: root.into(),
                 custom_main_file_stems,
                 settings,
+                builtin_plugins,
             };
 
             if compilation_unit.package == component.package {
@@ -240,6 +282,8 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
         let name = first_crate.name.clone();
         let discriminator = first_crate.discriminator.clone();
 
+        let builtin_plugins = crs.iter().flat_map(|cr| cr.builtin_plugins.clone()).collect();
+
         let custom_main_file_stems =
             crs.into_iter().flat_map(|cr| cr.custom_main_file_stems.unwrap()).collect();
 
@@ -249,6 +293,7 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
             root,
             custom_main_file_stems: Some(custom_main_file_stems),
             settings,
+            builtin_plugins,
         });
     }
 
@@ -347,4 +392,22 @@ fn scarb_package_experimental_features(
         associated_item_constraints: contains("associated_item_constraints"),
         coupons: contains("coupons"),
     }
+}
+
+/// Returns all plugins required by the `core` crate.
+fn plugins_for_corelib() -> Vec<BuiltinPlugin> {
+    vec![BuiltinPlugin::CairoTest, BuiltinPlugin::Starknet]
+}
+
+/// Returns all built-in plugins described by `dependencies`.
+fn plugins_from_dependencies(
+    metadata: &Metadata,
+    dependencies: &[&CompilationUnitCairoPluginMetadata],
+) -> Vec<BuiltinPlugin> {
+    dependencies
+        .iter()
+        .filter_map(|plugin_metadata| {
+            BuiltinPlugin::from_plugin_metadata(metadata, plugin_metadata)
+        })
+        .collect()
 }


src/project/scarb.rs
```diff
@@ -180,12 +193,36 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
...
+                .filter_map(|plugin_metadata| {
+                    let is_builtin = metadata.is_builtin(plugin_metadata);
+                    if !is_builtin.unwrap_or_default() {
+                        return None;
+                    }
+
+                    let builtin_plugin = BuiltinPlugin::try_new(plugin_metadata);
+                    Some(builtin_plugin?.suite())
+                })
+                .chain([get_default_plugin_suite()].into_iter())
```
Draggu, line 211: `get_default_plugin_suite()` should be added globally and ignored later

src/project/scarb.rs
```diff
@@ -240,12 +277,15 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
         let custom_main_file_stems =
             crs.into_iter().flat_map(|cr| cr.custom_main_file_stems.unwrap()).collect();
 
+        let plugins = get_default_plugin_suite();
```
Draggu, line 280: ?

integraledelebesgue: Funny, that's an outdated code

integraledelebesgue: My bad :v I changed this to a copy of the plugin suite assigned to the first crate in the group

src/project/scarb.rs
```diff
@@ -180,12 +193,33 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
...
+
+                    Some(builtin_plugin.suite())
+                })
+                .chain([get_default_plugin_suite()].into_iter())
+                .fold(PluginSuite::default(), |mut acc, suite| {
+                    acc.add(suite);
+                    acc
+                });
+
+            debug!("Plugins for {}: {:#?}", package.unwrap().id.repr, plugins);
```
piotmag769, line 214: Crates are all logged to debug here https://github.com/software-mansion/cairols/blob/7e357b34cf37c78ef2a428ab719484ac8ff0dc37/src/project/mod.rs#L95

src/project/scarb.rs
piotmag769: Imo it's better to add them here explicitly, we need them in `plugins` field for consistency with `ViewAnalyzedCrates` (we want to display active plugins there too)

src/project/crate_data.rs
```diff
@@ -79,7 +86,10 @@ impl Crate {
 
         let custom_main_file_stems = extract_custom_file_stems(db, crate_id);
 
-        Some(Self { name, discriminator, root, custom_main_file_stems, settings })
+        // TODO (later in the stack): Intern the suite and set as an override.
```
piotmag769, line 89: This comment is incorrect

src/project/scarb.rs
```diff
@@ -64,6 +67,16 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
...
         for component in &compilation_unit.components {
+            // Plugin components don't have crates associated with them.
+            if component
+                .id
+                .as_ref()
+                .and_then(|component_id| metadata.get_plugin(component_id))
+                .is_some()
+            {
+                continue;
+            }
```
piotmag769, line 78: This will never happen, plugins are not components

piotmag769 (changes requested):

src/project/scarb.rs
integraledelebesgue: You're right, fixed

src/project/scarb.rs
integraledelebesgue: This message is for my debugging purposes. I added a comment reminding me to delete it before merging

src/project/crate_data.rs
integraledelebesgue: Fixed :)

src/project/crate_data.rs
```diff
@@ -36,6 +37,10 @@ pub struct Crate {
 
     /// Crate settings.
     pub settings: CrateSettings,
+
+    /// Plugins required by the crate.
+    #[allow(dead_code)] // TODO: remove. The field is used later in the stack
+    pub plugins: PluginSuite,
```
piotmag769, line 43:
```suggestion
    /// Built-in plugins required by the crate.
    #[allow(dead_code)] // TODO: remove. The field is used later in the stack
    pub built_in_plugins: PluginSuite,
```

src/project/crate_data.rs
```diff
@@ -7,13 +7,14 @@ use cairo_lang_filesystem::db::{
...
 use cairo_lang_filesystem::ids::{CrateId, CrateLongId, Directory};
+use cairo_lang_semantic::plugin::PluginSuite;
 use cairo_lang_utils::{Intern, LookupIntern};
 use smol_str::SmolStr;
 
 use crate::lang::db::AnalysisDatabase;
 
 /// A complete set of information needed to set up a real crate in the analysis database.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
```
mkaput, line 17: I would like to preserve the plain nature of this struct. I know we're not based on upstream changes to the compiler yet here, but do you think if it would be possible to keep just a set of plugin ids here?

piotmag769: You don't have plugin ids in this PR at all since they are never interned, only treated as an input

integraledelebesgue: Storing IDs in `Crate` seems possible when we start using the new compiler API, but it would require passing a database to `extract_crates`, which can be tricky

mkaput: @piotmag769 is right. we would need to have a custom ID mechanism (PackageIDs I guess) at this stage. nonetheless, this is a separate piece of work which is not needed urgently right now

src/project/scarb.rs
integraledelebesgue: Done

src/project/crate_data.rs
integraledelebesgue: Done

src/project/scarb.rs
```diff
@@ -183,12 +182,47 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
...
+            let builtin_plugins = plugin_dependencies
+                .into_iter()
+                .filter_map(|plugin_metadata| {
+                    let builtin_plugin =
+                        BuiltinPlugin::from_plugin_metadata(metadata, plugin_metadata)?;
+
+                    Some(builtin_plugin.suite())
+                })
+                .fold(
+                    get_default_plugin_suite(), // All crates have basic plugins set by default.
```
piotmag769, line 212: You need to add tricks here too :c

src/project/crate_data.rs
```diff
@@ -36,6 +37,10 @@ pub struct Crate {
 
     /// Crate settings.
     pub settings: CrateSettings,
+
+    /// Built-in plugins required by the crate.
+    #[allow(dead_code)] // TODO: remove. The field is used later in the stack
+    pub builtin_plugins: PluginSuite,
```
piotmag769, line 43: We probably want to expose them in `ViewAnalyzedCrates` (the names of builtin plugins that were loaded), wdyt?

src/project/scarb.rs
integraledelebesgue: Done, all crates now have `Tricks`

src/project/crate_data.rs
piotmag769: Make an issue for that pls

src/project/scarb.rs
```diff
@@ -348,3 +387,46 @@ fn scarb_package_experimental_features(
...
+        .extra_plugin_suites
+        .map(|provider| provider())
+        .unwrap_or_default()
+}
+
+/// Returns all plugins required by the `core` crate.
+fn plugin_suite_for_corelib() -> PluginSuite {
+    [get_default_plugin_suite(), test_plugin_suite(), test_assert_suite()]
+        .into_iter()
+        .chain(tricks()) // All crates should receive the `Tricks`.
```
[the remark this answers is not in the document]
piotmag769: Not sure either, but it did until now xD

src/project/scarb.rs
```diff
@@ -348,3 +387,46 @@ fn scarb_package_experimental_features(
...
+/// Returns all built-in plugins described by `dependencies`.
+fn plugin_suite_from_dependencies(
+    metadata: &Metadata,
+    dependencies: &[&CompilationUnitCairoPluginMetadata],
+) -> PluginSuite {
+    dependencies
+        .iter()
+        .filter_map(|plugin_metadata| {
+            let builtin_plugin =
+                BuiltinPlugin::from_plugin_metadata(metadata, plugin_metadata)?;
```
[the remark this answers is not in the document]
integraledelebesgue: Not really, we could write a function that returns a `PluginSuite` based on the plugin metadata. But using an intermediate type with named variants seems *smoother* to me

src/project/scarb.rs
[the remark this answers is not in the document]
integraledelebesgue: That's a good question ;v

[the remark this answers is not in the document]
integraledelebesgue: We can remove `Tricks` from the corelib suite. If someone from Dojo complains, which is very unlikely, we will simply add them back

src/project/scarb.rs
```diff
@@ -183,12 +190,41 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
...
 
+            let plugin_dependencies = component
+                .dependencies
+                .as_ref()
+                .map(|dependencies| {
+                    dependencies
+                        .iter()
+                        .filter_map(|dependency| {
+                            compilation_unit.cairo_plugins.iter().find(|plugin| {
+                                match plugin.component_dependency_id {
```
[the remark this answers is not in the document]
integraledelebesgue: We can even do:
```rust
plugin.component_dependency_id.as_ref() == Some(&dependency.id)
```

instead of 

```rust
plugin
   .component_dependency_id
   .as_ref()
   .is_some_and(|id| id == &dependency.id)
```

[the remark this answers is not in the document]
integraledelebesgue: @piotmag769 suggested to add the `error!` log in the second branch: https://github.com/software-mansion/cairols/pull/209#discussion_r1952444957. I think I will stick to the `match` then

src/project/scarb.rs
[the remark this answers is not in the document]
piotmag769: Tbh no need to change it imo. We will get rid of `TRICKS` soon anyway

src/project/scarb.rs
```diff
@@ -183,12 +190,41 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
 
             let custom_main_file_stems = (file_stem != "lib").then_some(vec![file_stem.into()]);
 
+            let plugin_dependencies = component
```
piotmag769, line 193: Aggregate as discussed previously

src/project/scarb.rs
```diff
@@ -183,12 +184,54 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
...
+            // Procedural macros are handled separately in the
+            // `crate::lang::proc_macros::controller`.
+
+            let mut builtin_plugins = crates_by_component_id
+                .get(&component_id)
+                .as_ref()
+                .map(|cr| cr.builtin_plugins.clone())
+                .unwrap_or_default();
+
+            builtin_plugins.extend(if package.map(|p| p.name == "core").unwrap_or_default() {
```
Draggu, line 221:
```suggestion
            builtin_plugins.extend(if package.is_some_and(|p| p.name == "core") {
```

Draggu (approved):

src/project/scarb.rs
```diff
@@ -348,3 +394,26 @@ fn scarb_package_experimental_features(
         coupons: contains("coupons"),
     }
 }
+
+/// Returns all plugins required by the `core` crate.
+fn plugins_for_corelib() -> Vec<BuiltinPlugin> {
+    vec![BuiltinPlugin::CairoTest]
+}
```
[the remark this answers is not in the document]
integraledelebesgue: Sure, we can place it here. Done

src/project/scarb.rs
integraledelebesgue: Done

src/project/scarb.rs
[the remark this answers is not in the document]
piotmag769: #327

piotmag769 (approved):

src/project/scarb.rs
```diff
@@ -183,12 +184,53 @@ pub fn extract_crates(metadata: &Metadata) -> Vec<Crate> {
...
+                    dependencies
+                        .iter()
+                        .filter_map(|dependency| {
+                            let plugin = compilation_unit.cairo_plugins.iter().find(|plugin| {
+                                plugin.component_dependency_id.as_ref() == Some(&dependency.id)
+                            });
+
+                            if plugin.is_none() {
+                                error!(
+                                    "Plugin dependency `{}` not found in Scarb metadata",
```
mkaput, line 200:
```suggestion
                                    "plugin dependency not found in Scarb metadata: {}",
```

mkaput (approved):

src/project/scarb.rs
integraledelebesgue: This error has been deleted by @piotmag769 since it's not very informative. `plugin` being `None` usually means that the `dependency` is not a plugin and should be skipped. I must have messed up with the commits when we were updating the stack concurrently :)

PR #212, opened 2025-01-23
Migrate diagnostics testing to new notifications 

Arcticae:
Stack: 
⬇️ #44

### DIFF
--- a/tests/e2e/support/mock_client.rs
+++ b/tests/e2e/support/mock_client.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
 use std::ffi::OsStr;
 use std::path::Path;
 use std::time::Duration;
@@ -8,7 +8,7 @@ use cairo_language_server::build_service_for_e2e_tests;
 use cairo_language_server::lsp::ext::testing::ProjectUpdatingFinished;
 use lsp_server::{Message, Notification, Request, Response};
 use lsp_types::request::{RegisterCapability, Request as LspRequest};
-use lsp_types::{lsp_notification, lsp_request};
+use lsp_types::{Diagnostic, PublishDiagnosticsParams, Url, lsp_notification, lsp_request};
 use serde_json::Value;
 
 use crate::support::fixture::Fixture;
@@ -24,6 +24,8 @@ use crate::support::jsonrpc::RequestIdGenerator;
 /// blocking tasks are given a small period of time to complete.
 pub struct MockClient {
     fixture: Fixture,
+    // Keeps last diagnostics generation for each file
+    diagnostics: HashMap<Url, Vec<Diagnostic>>,
     req_id: RequestIdGenerator,
     client: lsp_server::Connection,
     trace: Vec<Message>,
@@ -52,6 +54,7 @@ impl MockClient {
             trace: Vec::new(),
             workspace_configuration,
             expect_request_handlers: Default::default(),
+            diagnostics: Default::default(),
         };
 
         std::thread::spawn(|| init().run_for_tests());
@@ -162,6 +165,12 @@ impl MockClient {
     pub fn trace(&self) -> &[Message] {
         &self.trace
     }
+
+    /// Gets state of diagnostics from previous round for the given file
+    pub fn get_diagnostics_for_file(&self, path: impl AsRef<Path>) -> Vec<Diagnostic> {
+        let url = self.fixture.file_url(&path);
+        self.diagnostics.get(&url).cloned().unwrap_or_default()
+    }
 }
 
 #[derive(Debug)]
@@ -187,6 +196,20 @@ impl MockClient {
             if let Message::Request(request) = &message {
                 if request.method == <lsp_request!("workspace/configuration")>::METHOD {
                     self.auto_respond_to_workspace_configuration_request(request);
+                }
+            }
+            if let Message::Notification(Notification { method, params }) = &message {
+                match method.as_str() {
+                    "textDocument/publishDiagnostics" => {
+                        let params: PublishDiagnosticsParams =
+                            serde_json::from_value(params.clone())
+                                .expect("failed to parse `textDocument/publishDiagnostics` params");
+                        self.diagnostics.insert(params.uri, params.diagnostics);
+                    }
+                    "cairo/diagnosticsCalculationStart" => {
+                        self.diagnostics.clear();
+                    }
+                    _ => {}
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -97,10 +97,10 @@ impl Config {
                     response.pop_front().as_ref().and_then(Value::as_bool).unwrap_or_default();
                 state.config.enable_proc_macros =
                     response.pop_front().as_ref().and_then(Value::as_bool).unwrap_or(false);
-
```
Draggu, line 100: newline

src/lang/proc_macros/controller.rs
```diff
@@ -187,7 +187,6 @@ impl ProcMacroClientController {
                 );
 
                 client.start_initialize();
-
```
Draggu, line 190: newline

tests/e2e/support/mock_client.rs
```diff
@@ -162,6 +165,12 @@ impl MockClient {
     pub fn trace(&self) -> &[Message] {
         &self.trace
     }
+
+    /// Gets state of diagnostics from previous round for the given file
+    pub fn get_diagnostics_for_file(&self, path: impl AsRef<Path>) -> Vec<Diagnostic> {
+        let url = self.fixture.file_url(&path);
+        self.diagnostics.get(&url).unwrap_or(&vec![]).clone()
```
Draggu, line 172:
```suggestion
        self.diagnostics.get(&url).cloned().unwrap_or_default()
```

tests/e2e/support/mock_client.rs
```diff
@@ -244,6 +253,11 @@ impl MockClient {
             predicate(&params).then_some(params)
         })
     }
+
+    // Remove messages from the trace that satisfy the given predicate
+    fn remove_messages_from_trace(&mut self, predicate: impl Fn(&Message) -> bool) {
+        self.trace = self.trace.clone().into_iter().filter(|a| !predicate(a)).collect()
```
Draggu, line 259:
```suggestion
        self.trace = self.trace.iter().filter(|a| !predicate(a)).cloned().collect()
```

tests/e2e/support/mock_client.rs
```diff
@@ -332,26 +346,39 @@ impl MockClient {
...
+    pub fn wait_for_diagnostics_round(&mut self) {
+        self.wait_for_notification::<DiagnosticsCalculationStart>(|_| true);
 
+        let mut diagnostics_from_round = HashMap::new();
+        while let Some(message) =
+            self.recv().unwrap_or_else(|err| panic!("{err:?}: Waiting for diagnostics"))
+        {
+            if let Message::Notification(Notification { method, params }) = message {
+                match method.as_str() {
+                    "textDocument/publishDiagnostics" => {
```
Draggu, line 358: Shouldn't we catch it in other cases too?

Arcticae: Discussed offline, resolving

tests/e2e/support/mock_client.rs
```diff
@@ -244,6 +267,66 @@ impl MockClient {
...
+                            seq += 1;
+                            advanced = true;
+                        }
+                    }
+                }
+                advanced
+            }};
+        }
+
+        macro_rules! check_sequence_end {
```
Draggu, line 289: inline

tests/e2e/support/mock_client.rs
```diff
@@ -244,6 +267,66 @@ impl MockClient {
             predicate(&params).then_some(params)
         })
     }
+
+    fn wait_for_notification_sequence(&mut self, notifications: Vec<&str>) {
+        let mut seq = 0;
+        // Block which checks if the notification matches the next expected one in the sequence
+        macro_rules! try_advance_sequence {
```
Draggu, line 274: make it a function

```
fn try_advance_sequence(message: &Message, seq: &mut usize) -> bool;
```

Draggu (changes requested):

Draggu (approved):

tests/e2e/support/mock_client.rs
```diff
@@ -162,6 +165,12 @@ impl MockClient {
     pub fn trace(&self) -> &[Message] {
         &self.trace
     }
+
+    /// Gets state of diagnostics from previous round for the given file
```
mkaput, line 169:
```suggestion
    /// Gets last received (if any) collection of diagnostics for given file.
```

tests/e2e/support/mock_client.rs
```diff
@@ -189,6 +198,20 @@ impl MockClient {
...
+                    "textDocument/publishDiagnostics" => {
+                        let params: PublishDiagnosticsParams =
+                            serde_json::from_value(params.clone())
+                                .expect("failed to parse `textDocument/publishDiagnostics` params");
+                        self.diagnostics.insert(params.uri, params.diagnostics);
+                    }
+                    "cairo/diagnosticsCalculationStart" => {
+                        self.diagnostics.clear();
+                    }
+                    _ => {}
```
mkaput, line 212: why are you doing this here? this is unlike the behaviour that editors seem to do, isn't it?

tests/e2e/support/mock_client.rs
```diff
@@ -332,26 +346,39 @@ impl MockClient {
...
-        &mut self,
-        path: impl AsRef<Path>,
-    ) -> lsp_types::PublishDiagnosticsParams {
+    /// `cairo/projectUpdatingFinished` and the next diagnostics round,
+    /// denoted by `cairo/diagnosticsCalculationStart` and `cairo/diagnosticsCalculationFinish`.
+    pub fn open_and_wait_for_diagnostics(&mut self, path: impl AsRef<Path>) -> Vec<Diagnostic> {
         let path = path.as_ref();
         self.open_and_wait_for_project_update(path);
-        self.wait_for_diagnostics(path)
+        self.clear_diagnostics_events_from_trace();
```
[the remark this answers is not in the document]
mkaput: I cannot get it why you are doing all these clears here. What was wrong with keeping the last known state of diagnostics, and just track whether something is being currently calculated or not? Have you tried this approach?

mkaput (changes requested):

tests/e2e/support/mock_client.rs
Arcticae: I keep only last diagnostics generation to avoid maintaining complicated logic. We don't need that behavior in the tests anyway.

tests/e2e/support/mock_client.rs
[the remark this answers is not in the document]
Arcticae: Because i want to catch a specific generations' diagnostics, the one that STARTS and ENDS after project update specifically (having both assessed is important here for correctness). I want the order of notifications to be correct.

I can also try changing the approach to have a more elaborate "assert" for diagnostics with timeout which waits for diagnostics with a timer, but that approach seems less reliable to me.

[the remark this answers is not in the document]
Arcticae: Let's consider a series of events:

1. You load up the project
2. The diagnostics start calculating  (receive `cairo/diagnosticsCalculationStart`)
3. You open up a file (receive `file/didOpen`)
4. Project is updated (receive `cairo/projectUpdatingFinished`)
5. The diagnostics are cancelled, or finished (receive `cairo/diagnosticsCalculationFinish` or not - not really deterministic)
6. In the test you wait for notification `Start`
7. receive `cairo/diagnosticsCalculationStart` (caused by previous `cairo/projectUpdatingFinished` from step 4)
8. Start received, wait for `Finish`-> return immediately because it's in the trace (step 5 or not - not really deterministic). 

At this point you have inconsistent diagnostics because you asserted the wrong generation. That's why asserting with preserving order is important in the tests (`file/didOpen` -> `cairo/projectUpdatingFinished` -> causes `cairo/diagnosticsCalculationStart` + `cairo/diagnosticsCalculationFinish` sequence). 

This can be achieved via: 
- Asserting if we have the correct sequence 
- Clearing previous ocurrences accordingly 

The solution implemented is a hybrid of both for preserving the existing APIs (`open_and_wait_for_diagnostics`, `open_and_wait_for_project_update`)

What i can do here at most is avoid clearing those notifications and assert that the order is correct here. Although that kind of behavior is vulnerable to changes of notification sending order in the LS.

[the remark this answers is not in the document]
mkaput: discussing this offline

mkaput (approved):

piotmag769 (commented):
dots, otherwise weird but G

PR #215, opened 2025-01-24
Add autocompletion for module name

Draggu:
This PR adds autocompletions for

```cairo

mod a<caret>

```

based on already existing files.

### DIFF
--- a/src/ide/completion/mod_.rs
+++ b/src/ide/completion/mod_.rs
@@ -1,0 +1,169 @@
+use std::collections::HashSet;
+use std::fs::DirEntry;
+use std::path::{Path, PathBuf};
+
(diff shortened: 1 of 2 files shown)


src/ide/completion/mod_.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+fn already_typed_text(
+    db: &AnalysisDatabase,
+    module: ItemModule,
+    node: SyntaxNode,
+) -> Option<String> {
+    if module.module_kw(db).as_syntax_node() == node {
+        Some(String::new())
+    } else {
```
piotmag769, line 101: Can we add a testcase for that (can be in a next PR)?

piotmag769 (approved):

PR #241, opened 2025-02-01
Refactor SyntaxNodeExt

Draggu:
**Stack**:
- #247
- #246
- #245
- #244
- #243
- #242
- #241 ⬅

### DIFF
--- a/src/lang/syntax.rs
+++ b/src/lang/syntax.rs
@@ -26,14 +26,22 @@ pub trait SyntaxNodeExt {
     /// Checks whether this syntax node is or is under the given syntax node in the syntax tree.
     fn is_descendant_or_self(&self, node: &SyntaxNode) -> bool;
 
-    /// Finds the first parent of a given kind.
+    /// Finds the first ancestor of a given kind.
+    fn ancestor_of_kind(&self, db: &dyn SyntaxGroup, kind: SyntaxKind) -> Option<SyntaxNode>;
+
+    /// Finds the first ancestor of a given kind and returns it in typed form.
+    fn ancestor_of_type<T: TypedSyntaxNode>(&self, db: &dyn SyntaxGroup) -> Option<T>;
+
+    /// Finds the parent of a given kind.
+    #[allow(dead_code)]
     fn parent_of_kind(&self, db: &dyn SyntaxGroup, kind: SyntaxKind) -> Option<SyntaxNode>;
 
-    /// Finds the first parent of a given kind and returns it in typed form.
+    /// Finds the parent of a given kind and returns it in typed form.
+    #[allow(dead_code)]
     fn parent_of_type<T: TypedSyntaxNode>(&self, db: &dyn SyntaxGroup) -> Option<T>;
 
     /// Finds the first parent of one of the kinds.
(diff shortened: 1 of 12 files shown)


Draggu:
Please review some of the following PRs first to see what this macro does

mkaput (changes requested):
please rewrite this stack as discussed offline

src/lang/syntax.rs
```diff
@@ -23,14 +23,41 @@ pub trait SyntaxNodeExt {
...
+
+    /// Finds the first ancestor of a given kind.
+    /// Will return `None` if node of any provided kind is found first.
+    #[allow(dead_code)]
+    fn ancestor_of_kind_failing_on(
+        &self,
+        db: &dyn SyntaxGroup,
+        kind: SyntaxKind,
+        kinds: &[SyntaxKind],
+    ) -> Option<SyntaxNode>;
```
piotmag769, line 57: "yes, I understand this is a bit absurd but that’s what we have"

piotmag769 (approved):

src/lang/syntax.rs
mkaput, line 57: I've looked up where you're needing these methods and I don't get why you need the failing behaviour. can you explain please?

mkaput (changes requested):

src/lang/syntax.rs
Draggu: I want to check if caret is in top level module, ie

```
mod a {
    mod b<should NOT work>;
}

mod c<should work>;
```

now thinking about this we can achieve it with simple negation, but I believe it can be useful in other cases too.

mkaput: I contest, because I find this method highly unreadable and doing these filters in-line would be easier to grasp

src/lang/syntax.rs
```diff
@@ -26,14 +26,22 @@ pub trait SyntaxNodeExt {
...
 
-    /// Finds the first parent of a given kind.
+    /// Finds the first ancestor of a given kind.
+    fn ancestor_of_kind(&self, db: &dyn SyntaxGroup, kind: SyntaxKind) -> Option<SyntaxNode>;
+
+    /// Finds the first ancestor of a given kind and returns it in typed form.
+    fn ancestor_of_type<T: TypedSyntaxNode>(&self, db: &dyn SyntaxGroup) -> Option<T>;
+
+    /// Finds the parent of a given kind.
+    #[allow(dead_code)]
```
mkaput, line 36:
```suggestion
    #[expect(dead_code)]
```

src/lang/syntax.rs
```diff
@@ -26,14 +26,22 @@ pub trait SyntaxNodeExt {
...
     fn parent_of_kind(&self, db: &dyn SyntaxGroup, kind: SyntaxKind) -> Option<SyntaxNode>;
 
-    /// Finds the first parent of a given kind and returns it in typed form.
+    /// Finds the parent of a given kind and returns it in typed form.
+    #[allow(dead_code)]
```
mkaput, line 40:
```suggestion
    #[expect(dead_code)]
```

mkaput (approved):

src/lang/syntax.rs
Draggu: Removed later anyway

PR #258, opened 2025-02-05
ViewSyntaxTree: branch tree to cursor

piotmag769:


Cargo.toml
```diff
@@ -45,8 +45,10 @@ cairo-lang-project = "*"
 cairo-lang-semantic = "*"
 cairo-lang-starknet = "*"
 cairo-lang-syntax = "*"
+cairo-lang-syntax-codegen = "*"
```
mkaput, line 48: missing patch

mkaput (approved):

Cargo.toml
piotmag769: It was already there though

mkaput: ah indeed, sorry for confusion

Arcticae (approved):

PR #260, opened 2025-02-05
Property initialize ItemDef inside MemberDef and VariantDef

mkaput:
We were feeding member/variant names as definition nodes for their items here... This has been slipping through because nothing relied on these further.

---

**Stack**:
- #267
- #261
- #260 ⬅

### DIFF
--- a/src/lang/defs.rs
+++ b/src/lang/defs.rs
@@ -338,12 +338,9 @@ impl MemberDef {
         member_id: MemberId,
         definition_node: SyntaxNode,
     ) -> Option<Self> {
-        let structure = ItemDef::new(db, &definition_node)?;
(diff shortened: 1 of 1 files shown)


src/lang/defs.rs
```diff
@@ -338,12 +338,9 @@ impl MemberDef {
         member_id: MemberId,
         definition_node: SyntaxNode,
     ) -> Option<Self> {
-        let structure = ItemDef::new(db, &definition_node)?;
```
piotmag769, line 341: It worked bcs `ItemDef` collects lookup items. Anyways your approach is better

piotmag769 (approved):

Draggu (approved):

PR #270, opened 2025-02-07
Track proc macro server status in analysis progress

Arcticae:
Stack: 
⬆️ #271

PR #271, opened 2025-02-07
Test diagnostics on custom proc macros

Arcticae:
Stack:
⬇️  #270

### DIFF
--- a/tests/e2e/custom_macros/mod.rs
+++ b/tests/e2e/custom_macros/mod.rs
@@ -15,7 +15,7 @@ struct DiagnosticsWithUrl {
 
 #[derive(Serialize)]
 struct DiagnosticsReport {
-    diagnostics: String,
+    diagnostics: Vec<DiagnosticsWithUrl>,
 }
 
 #[test]
@@ -79,11 +79,8 @@ fn test_custom_macro() {
         .into_iter()
         .filter(|(url, _)| !url.path().contains("core/src"))
         .sorted_by(|(url_a, _), (url_b, _)| url_a.path().cmp(url_b.path()))
-        .map(|(url, diagnostics)| DiagnosticsWithUrl { url: url.to_string(), diagnostics })
+        .map(|(url, diagnostics)| DiagnosticsWithUrl { url: normalize(&ls, url), diagnostics })
         .collect();
 
-    let serialized_diags = serde_json::to_string_pretty(&sorted_diagnostics).unwrap();
-    let normalized_diags = normalize(&ls, serialized_diags);
-
-    insta::assert_toml_snapshot!(DiagnosticsReport { diagnostics: normalized_diags })
+    insta::assert_json_snapshot!(DiagnosticsReport { diagnostics: sorted_diagnostics })
 }
--- a/tests/e2e/custom_macros/mod.rs
+++ b/tests/e2e/custom_macros/mod.rs
@@ -1,0 +1,89 @@
+use indoc::indoc;
+use itertools::Itertools;
+use lsp_types::Diagnostic;
+use serde::Serialize;
+use serde_json::json;
+
+use crate::support::normalize::normalize;
+use crate::support::sandbox;
+
+#[derive(Serialize)]
+struct DiagnosticsWithUrl {
+    url: String,
+    diagnostics: Vec<Diagnostic>,
+}
+
+#[derive(Serialize)]
+struct DiagnosticsReport {
+    diagnostics: String,
+}
+
+#[test]
+fn test_custom_macro() {
+    let mut ls = sandbox! {
+        files {
+            "a/Scarb.toml" => indoc! (r#"
+                [package]
+                name = "a"
+                version = "0.1.0"
+                edition = "2024_07"
+                [dependencies]
+                macros = { path = "../macros" }
+            "#),
+            "a/src/lib.cairo" => indoc!(r#"
+                #[decorate]
+                fn decorated() {}
+            "#),
+            "macros/Cargo.toml" => indoc!(r#"
+                [package]
+                name = "some_macro"
+                version = "0.1.0"
+                edition = "2021"
+                publish = false
+
+                [lib]
+                crate-type = ["rlib", "cdylib"]
+
+                [dependencies]
+                cairo-lang-macro = "0.1.1"
+                cairo-lang-parser = "2.7.0"
+
+            "#),
+            "macros/Scarb.toml" => indoc!(r#"
+                [package]
+                name = "macros"
+                version = "0.1.0"
+
+                [cairo-plugin]
+            "#),
+            "macros/src/lib.rs" => indoc!(r#"
+                use cairo_lang_macro::{ProcMacroResult, TokenStream, attribute_macro};
+
+                #[attribute_macro]
+                pub fn decorate(_args: TokenStream, item: TokenStream) -> ProcMacroResult {
+                    let new_res = format!("{} fn added_fun() {{ a = b; }}", item); // Syntax error
+                    ProcMacroResult::new(TokenStream::new(new_res))
+                }
+            "#),
+        }
+        cwd = "/a";
+        workspace_configuration = json!({
+            "cairo1": {
+                "enableProcMacros": true,
+            }
+        });
+    };
+
+    let newest_diagnostics = ls.open_and_wait_for_diagnostics_generation("a/src/lib.cairo");
+    let sorted_diagnostics: Vec<DiagnosticsWithUrl> = newest_diagnostics
+        .into_iter()
+        .filter(|(url, _)| !url.path().contains("core/src"))
+        .sorted_by(|(url_a, _), (url_b, _)| url_a.path().cmp(url_b.path()))
+        .map(|(url, diagnostics)| DiagnosticsWithUrl { url: url.to_string(), diagnostics })
+        .collect();
+
+    let serialized_diags = serde_json::to_string_pretty(&sorted_diagnostics).unwrap();
+    let normalized_diags = normalize(&ls, serialized_diags);
+
+    insta::assert_toml_snapshot!(DiagnosticsReport { diagnostics: normalized_diags })
+}
--- a/tests/e2e/custom_macros/mod.rs
+++ b/tests/e2e/custom_macros/mod.rs
@@ -22,19 +22,19 @@ struct DiagnosticsReport {
 fn test_custom_macro() {
     let mut ls = sandbox! {
         files {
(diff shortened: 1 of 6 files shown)


tests/e2e/custom_macros/snapshots/e2e__custom_macros__custom_macro.snap
```diff
@@ -0,0 +1,55 @@
+---
+source: tests/e2e/custom_macros/mod.rs
+expression: "DiagnosticsReport { diagnostics: normalized_diags }"
+snapshot_kind: text
+---
+diagnostics = '''
+[
+  {
+    "url": "[ROOT_URL]a/src/lib.cairo",
+    "diagnostics": [
```
Draggu, line 10: Can we map them to the original line code? It is hard to tell where it refers to.

tests/e2e/support/mod.rs
```diff
@@ -34,6 +35,13 @@ macro_rules! sandbox {
 
         $($(fixture.add_file($file, $content);)*)?
 
+        $(
+            use std::env;
+            let mut p = fixture.root_path().into_os_string();
+            p.push($cwd);
+            env::set_current_dir(p).unwrap();
```
Draggu, line 42:
```suggestion
            let mut p = fixture.root_path().into_os_string();
            p.push($cwd);
            ::std::env::set_current_dir(p).unwrap();
```

tests/e2e/support/mock_client.rs
```diff
@@ -428,6 +440,74 @@ impl MockClient {
...
+
+            StreamFlowControl::Continue
+        }));
+
+        // It should contain all accumulated diagnostics up to the end of the generation
+        self.diagnostics.clone()
+    }
+
+    fn subscribe_notifications(&mut self, stream_consumer: StreamConsumer<Notification>) {
+        let invoke_consumer = |msg: &Message| {
```
Draggu, line 488:
```suggestion
        let consume_notification = |msg: &Message| {
```

tests/e2e/support/mock_client.rs
```diff
@@ -428,6 +440,74 @@ impl MockClient {
...
+                    AnalysisFinished => {
+                        if *in_progress_open && *project_updated {
+                            return StreamFlowControl::Stop;
+                        }
+                    }
+                }
+            }
+            if notification.method == "cairo/projectUpdatingFinished" {
+                *project_updated = true;
+            }
```
Draggu, line 478:
```suggestion
            *project_updated = notification.method == "cairo/projectUpdatingFinished";
```

tests/e2e/support/mock_client.rs
```diff
@@ -359,6 +363,14 @@ impl MockClient {
     }
 }
 
+#[derive(PartialEq)]
+enum StreamFlowControl {
```
Draggu, line 367: Probably `std::ops::ControlFlow` will make better, it can be `?` propagated too.

Draggu (changes requested):

tests/e2e/support/mock_client.rs
Draggu: Ignore this

tests/e2e/custom_macros/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    diagnostics: Vec<Diagnostic>,
+}
+
+#[derive(Serialize)]
+struct DiagnosticsReport {
+    diagnostics: String,
+}
+
+#[test]
+fn test_custom_macro() {
```
mkaput, line 22: I'm afraid our CI times will explode with this because of Rust compilation. I wonder if we could prepublish this macro to scarbs.xyz, with precompiled binaries and use it that way. Wdyt? No need to address this in this PR, but worth deciding on now

mkaput (changes requested):

tests/e2e/custom_macros/mod.rs
Arcticae: Yeah that's something that i want to address as a follow up. 
Let me create an issue

Arcticae: https://github.com/software-mansion/cairols/issues/274

tests/e2e/custom_macros/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+        .into_iter()
+        .filter(|(url, _)| !url.path().contains("core/src"))
+        .sorted_by(|(url_a, _), (url_b, _)| url_a.path().cmp(url_b.path()))
+        .map(|(url, diagnostics)| DiagnosticsWithUrl { url: url.to_string(), diagnostics })
+        .collect();
+
+    let serialized_diags = serde_json::to_string_pretty(&sorted_diagnostics).unwrap();
+    let normalized_diags = normalize(&ls, serialized_diags);
+
+    insta::assert_toml_snapshot!(DiagnosticsReport { diagnostics: normalized_diags })
```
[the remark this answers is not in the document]
Arcticae: TOML doesn't support nesting that i want to achieve with this, that's why im using json

tests/e2e/support/mock_client.rs
```diff
@@ -428,6 +440,74 @@ impl MockClient {
...
+            }
+
+            StreamFlowControl::Continue
+        }));
+
+        // It should contain all accumulated diagnostics up to the end of the generation
+        self.diagnostics.clone()
+    }
+
+    fn subscribe_notifications(&mut self, stream_consumer: StreamConsumer<Notification>) {
```
piotmag769, line 487: This name is misleading, you don't register any subscribtion here. This name doesn't even suggest that this function waits for sth

piotmag769 (commented):
Why cannot you do the way you did it before?

tests/e2e/support/mock_client.rs
Arcticae: I think this method is descriptive (and standard) enough 
Please suggest a better name here if you think it's bad

tests/e2e/custom_macros/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+                "enableProcMacros": true,
+            }
+        });
+    };
+
+    let newest_diagnostics = ls.open_and_wait_for_diagnostics_generation("a/src/lib.cairo");
+    let sorted_diagnostics: Vec<DiagnosticsWithUrl> = newest_diagnostics
+        .into_iter()
+        .filter(|(url, _)| !url.path().contains("core/src"))
+        .sorted_by(|(url_a, _), (url_b, _)| url_a.path().cmp(url_b.path()))
```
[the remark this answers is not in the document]
Arcticae: If there's more diagnostics we know something is off (and/or test was changed)

tests/e2e/custom_macros/mod.rs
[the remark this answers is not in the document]
Arcticae: i believe the problem is that there is a files list here and nested inside there's diags list. i don't think it's worth the time figuring out a way of doing it with the toml. do you?

tests/e2e/custom_macros/snapshots/e2e__custom_macros__custom_macro.snap
Arcticae: Do you want me to: 
1. Enable diagnostics tracing
2. Map each individual diagnostic to a related code line, and print out the whole (expanded) line here?

If you mean 2nd one, it's a bit work since i have to fetch the whole document for each diagnostic, expand that, and then get only the line i want 
Not sure it's worth doing ATM

tests/e2e/custom_macros/mod.rs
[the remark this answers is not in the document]
mkaput: you can just use json serializer that insta provides instead of toml. we already have `serde_json`, so enabling the `json` feature in `insta` is gonna be cheap

tests/e2e/custom_macros/snapshots/e2e__custom_macros__custom_macro.snap
Draggu: I mean 2. create issue at least

tests/e2e/support/mock_client.rs
```diff
@@ -428,6 +441,74 @@ impl MockClient {
...
+
+        // It should contain all accumulated diagnostics up to the end of the generation.
+        self.diagnostics.clone()
+    }
+
+    fn subscribe_notifications(&mut self, stream_consumer: StreamConsumer<Notification>) {
+        let consume_notification = |msg: &Message| {
+            if let Message::Notification(notification) = msg {
+                let flow_control = stream_consumer(notification.clone());
+                return flow_control == ControlFlow::Break(());
```
Draggu, line 492:
```suggestion
                stream_consumer(notification.clone())?;
```

tests/e2e/support/mock_client.rs
```diff
@@ -428,6 +441,74 @@ impl MockClient {
...
+            }
+            false
+        };
+
+        for msg in self.trace() {
+            let end_stream = consume_notification(msg);
+            if end_stream {
```
Draggu, line 499:
```suggestion
            if consume_notification(msg).is_break() {
```

tests/e2e/support/mock_client.rs
```diff
@@ -428,6 +441,74 @@ impl MockClient {
...
+                return;
+            }
+        }
+
+        while let Some(msg) = self.recv().expect("Cannot read from the server") {
+            let end_stream = consume_notification(&msg);
+            if end_stream {
```
Draggu, line 506: vide supra

mkaput (approved):
unlocking as I'm going OOO

tests/e2e/custom_macros/snapshots/e2e__custom_macros__custom_macro.snap
Arcticae: #292

Draggu (approved):

tests/e2e/custom_macros/mod.rs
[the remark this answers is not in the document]
piotmag769: I guess

tests/e2e/support/mock_client.rs
piotmag769: You could actually inline it to sth like 
```rs
let mut consume_notification = |msg: &Message| -> ControlFlow<()> {
            if let Message::Notification(notification) = msg {
                stream_consumer(notification.clone())? // <- Inline stream_consumer
            }
            ControlFlow::Continue(())
        };

// inline the rest
        
```

For me it's more natural, but I don't have a strong opinion, do as u think is best

PR #277, opened 2025-02-07
Fix panics in viewSyntaxTree

piotmag769:


src/ide/introspection/syntax_tree/printer.rs
```diff
@@ -70,16 +71,29 @@ pub fn syntax_tree_branch_above_leaf(
...
+
+            let eof_green =
+                children[SyntaxFile::INDEX_EOF].lookup_intern(db).children()[1].lookup_intern(db);
+            match &eof_green.details {
+                GreenNodeDetails::Token(text) => {
+                    printer.print_token_node("eof", "", "", text.clone(), eof_green.kind)
+                }
+                GreenNodeDetails::Node { .. } => unreachable!(),
+            }
+        }
```
piotmag769, line 96: This can be reached in case where the caret is placed AFTER the last module item in the file:
```cairo
fn uwu {

}<caret>

```
In this case `db.find_syntax_node_at_position` returns `SyntaxNode` of type `SyntaxFile` (I guess this happens due to some offset logic, haven't investigated it further). The most logical solution is to assume we are at EOF here (or its leading trivia, but I chose the former).

mkaput (approved):

src/server/routing/traits.rs
```diff
@@ -389,7 +389,11 @@ impl BackgroundDocumentRequestHandler for ViewSyntaxTree {
         _notifier: Notifier,
         params: TextDocumentPositionParams,
     ) -> LSPResult<Option<String>> {
-        Ok(ide::introspection::syntax_tree::get_syntax_tree_for_file(&snapshot.db, params))
+        Ok(if is_cairo_file_path(&params.text_document.uri) {
+            ide::introspection::syntax_tree::get_syntax_tree_for_file(&snapshot.db, params)
+        } else {
+            None
+        })
```
Draggu, line 396:
```suggestion
        Ok(if is_cairo_file_path(&params.text_document.uri).then(||ide::introspection::syntax_tree::get_syntax_tree_for_file(&snapshot.db, params)).flatten()
)
```

Draggu (approved):

PR #304, opened 2025-02-14
improve code_actions code

Draggu:
Also remove `.unwrap()` from previous PR

### DIFF
--- a/src/ide/code_actions/mod.rs
+++ b/src/ide/code_actions/mod.rs
@@ -1,5 +1,4 @@
 use cairo_lang_syntax::node::SyntaxNode;
-use cairo_lang_utils::ordered_hash_map::{Entry, OrderedHashMap};
 use itertools::Itertools;
 use lsp_types::{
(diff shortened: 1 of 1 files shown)


src/ide/code_actions/mod.rs
```diff
@@ -54,61 +53,59 @@ fn get_code_actions_for_diagnostics(
...
+    let Some(ctx) = AnalysisContext::from_node(db, node.clone()) else { return Default::default() };
 
     diagnostic_groups_by_codes
         .into_iter()
-        .flat_map(|(code, diagnostics)| match code.as_str() {
+        .flat_map(|(code, diagnostics)| match code {
             "E0001" => diagnostics
                 .into_iter()
-                .filter_map(|diagnostic| {
+                // There should be exactly one diagnostic.
```
piotmag769, line 70: How do you know that?

Draggu: We receives diagnostic that are in range for this request, and it is send for diagnostic. With different diagnostics there can be overlap but not of same kind.

integraledelebesgue (approved):

PR #305, opened 2025-02-14
Simple rename impl

piotmag769:


src/ide/navigation/rename.rs
```diff
@@ -1,7 +1,33 @@
...
-    None
+use crate::lang::db::{AnalysisDatabase, LsSyntaxGroup};
+use crate::lang::defs::SymbolDef;
+use crate::lang::lsp::{LsProtoGroup, ToCairo};
+
+// TODO: handle non-inline modules and crates separately (files need to be renamed too).
+pub fn rename(params: RenameParams, db: &AnalysisDatabase) -> Option<WorkspaceEdit> {
+    let new_name = params.new_name;
+
+    let is_new_name_valid = new_name.chars().all(|c| c.is_ascii_alphanumeric() || c == '_');
```
Draggu, line 13: comment where it is copied from

Draggu, line 13:
```suggestion
    let is_valid_ident = new_name.chars().all(|c| c.is_ascii_alphanumeric() || c == '_');
```

src/ide/navigation/references.rs
```diff
@@ -16,30 +14,5 @@ pub fn references(params: ReferenceParams, db: &AnalysisDatabase) -> Option<Vec<
...
-    .filter_map(|(file, span)| {
-        let found_uri = db.url_for_file(file)?;
-        let range = span.position_in_file(db.upcast(), file)?.to_lsp();
-        let location = Location { uri: found_uri, range };
-        Some(location)
-    })
-    .collect();
-
-    Some(locations)
+    Some(symbol.locations(db, include_declaration))
```
mkaput, line 17: I don't like this refactoring -- it extracts too many distinct tasks at once:

1. Joining declaration & usages.
2. Simplifying `FoundUsage` to a tuple.
3. Producing `Location` out of these tuples.

I have an alternative idea. Will make a PR out of this

mkaput: https://github.com/software-mansion/cairols/pull/314

piotmag769: Good idea, will reuse it when merged

src/lang/defs/mod.rs
```diff
@@ -170,4 +173,28 @@ impl SymbolDef {
...
         FindUsages::new(self, db)
     }
+
+    /// Finds all locations in which this symbol was used.
+    pub fn locations(&self, db: &AnalysisDatabase, include_declaration: bool) -> Vec<Location> {
+        // TODO(mkaput): Think about how to deal with `mod foo;` vs `mod foo { ... }`.
+        // Location where the searched symbol is declared.
+        // This can rarely be `None`, for example, for macros.
+        // For all cases we cover here, definition == declaration.
+        let declaration = self.definition_location(db).filter(|_| include_declaration);
```
Arcticae, line 183: This is a weird operation in itself, but i dislike the combination of those weird ops to achieve including the declaration or not, based on a boolean flag.

The logic doesn't seem to be needed in more than one place btw so it's not that bad to call 2 separate functions, or add some code in one place and leave it out from the other one

src/ide/navigation/rename.rs
```diff
@@ -1,7 +1,33 @@
...
+    if !is_new_name_valid {
+        return None;
```
Arcticae, line 15: How will the errors be handled here? It'd be nice to inform the user abt. the reasoning for aborting the rename from within an err tooltip

src/lang/defs/mod.rs
piotmag769: Not relevant anymore

src/ide/navigation/rename.rs
piotmag769: Done

piotmag769: image

src/ide/navigation/rename.rs
```diff
@@ -1,7 +1,51 @@
...
+use lsp_types::{Location, MessageType, RenameParams, ShowMessageParams, TextEdit, WorkspaceEdit};
 
-pub fn rename(_params: RenameParams, _db: &AnalysisDatabase) -> Option<WorkspaceEdit> {
-    None
+use crate::lang::db::{AnalysisDatabase, LsSyntaxGroup};
+use crate::lang::defs::SymbolDef;
+use crate::lang::lsp::{LsProtoGroup, ToCairo};
+use crate::server::client::Notifier;
+
+// TODO: handle non-inline modules and crates separately (files need to be renamed too).
```
Draggu, line 12: Please place a selector for these cases at least now and return None if it matches

Draggu (changes requested):

src/ide/navigation/rename.rs
piotmag769: Done

Draggu (approved):

mkaput (approved):

PR #311, opened 2025-02-17
Fix looking for scarb cache path

piotmag769:


Draggu (approved):

src/toolchain/scarb.rs
```diff
@@ -196,4 +203,21 @@ impl ScarbToolchain {
 
         Ok(version)
     }
+
+    fn fetch_cache_path(&self) -> Result<PathBuf> {
+        let Some(scarb_path) = self.discover() else { bail!("failed to get scarb path") };
+
+        let output = Command::new(scarb_path).arg("cache").arg("path").output()?;
```
maciektr, line 210: Why not use `SCARB_CACHE` if it's available? 🤔

integraledelebesgue (approved):

src/toolchain/scarb.rs
piotmag769: It is kinda an impl detail (even though it is guaranteed in docs). We still call this command once per LS instance (`OnceCell`) so it's fine imo

PR #352, opened 2025-02-24
Fix dynamic disabling lint

### LINKED ISSUE #350: Lint diagnostics are not disappearing when disabling lint in config
### Problem

See _steps_.

### Steps

1. Set `enableLinter` to `false`.
2. Open a project.
3. Wait for diagnostics to compute.
4. Set `enableLinter` to `true`.
5. Wait for lints to appear.
6. Set `enableLinter` to `false.

Lints should now disappear from the _Problems_ panel, but they don't.

### Possible Solution(s)

_No response_

### Notes

Surprisigly enabling linter in runtime works - lints diagnostics will be computed. It's only a problem to remove these diagnostics when disabling the linter.

### Scarb Version

```text

```

Draggu:
close #350

### DIFF
--- a/src/lang/plugins.rs
+++ b/src/lang/plugins.rs
@@ -2,7 +2,11 @@ use cairo_lang_semantic::plugin::AnalyzerPlugin;
 
 pub trait AnalyzerPluginType: AnalyzerPlugin {
     fn is_cairo_lint_plugin(&self) -> bool {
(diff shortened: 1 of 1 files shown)


mkaput (approved):

piotmag769 (commented):
Lol xD

PR #353, opened 2025-02-24
Fix inline procmacro expansion

### LINKED ISSUE #351: Inline procmacro expansion is not working on multiline & indented usage
### Problem

As title says, proc macro expansion seems to be broken in such cases. See _steps_.

### Steps

This doesn't expand properly: 

Image

This does expand: 

Image

### Possible Solution(s)

_No response_

### Notes

_No response_

### Scarb Version

```text

```

Draggu:
close #351


Arcticae (approved):

piotmag769 (commented):
:+1:

PR #356, opened 2025-02-25
`MockClient` request handling refactor

Arcticae:


piotmag769 (changes requested):
I think the way we should do it is how @mkaput proposed which is handling incoming requests in `recv` itself. Right now we still have a problem that there are a few `recv` calls, but only the usage that was modified in this PR handles requests from the server.

mkaput:
Naturally I agree with @piotmag769

tests/e2e/support/mock_client.rs
```diff
@@ -215,6 +193,11 @@ impl MockClient {
                 if request.method == <lsp_request!("workspace/configuration")>::METHOD {
                     self.auto_respond_to_workspace_configuration_request(request);
                 }
+                if let Some(handler) = self.get_handler_for(&request.method) {
+                    let response = (handler.f)(request);
+                    let message = Message::Response(response);
+                    self.client.sender.send(message).expect("failed to send response");
+                }
```
piotmag769, line 200: I think we should panic if there is not handler found. If a request from a server is ignored it may lead to some weird behaviors

Arcticae: Done

PR #363, opened 2025-02-25
Reduce the amount of parallelism consumed by our CI

mkaput:
- Merged `udeps` job into `check`,
- reduced the number of parallel test runs from 8 to 4,
- removed platform specs from matrices; we only run Linux anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,38 +14,29 @@ concurrency:
 
 jobs:
   build-test:
-    name: build test ${{ matrix.platform.name }}
-    runs-on: ${{ matrix.platform.os }}
-    strategy:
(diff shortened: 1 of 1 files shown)


mkaput:
before: https://github.com/software-mansion/cairols/actions/runs/13522775544
after: https://github.com/software-mansion/cairols/actions/runs/13524507963

<1 minute slower execution time, but uses 50% less parallel instances, which I expect to reduce delays we experience

Arcticae (approved):

Arcticae:
We need to update the required checks, LMK if you want me to do this

mkaput:
> We need to update the required checks, LMK if you want me to do this

I will update the branch rules myself

PR #366, opened 2025-02-25
Refresh semantic tokens when macros are known

### LINKED ISSUE #354: Semantic highlightning of proc macros is not auto refreshed
### Problem

When we have proc macros in our code, we color them as functions before these are loaded. This can be fixed by any file modification which is triggering a refresh of semantic diagnostics. It should be done automatically.
Image

### Steps

Open project with proc-macro and do nothing

### Possible Solution(s)

_No response_

### Notes

_No response_

### Scarb Version

```text

```

piotmag769:
Fixes #354

### DIFF
--- a/src/lang/proc_macros/controller.rs
+++ b/src/lang/proc_macros/controller.rs
@@ -15,7 +15,8 @@ use governor::clock::QuantaClock;
 use governor::state::{InMemoryState, NotKeyed};
 use governor::{Quota, RateLimiter};
 use lsp_types::notification::ShowMessage;
-use lsp_types::{MessageType, ShowMessageParams};
+use lsp_types::request::SemanticTokensRefresh;
+use lsp_types::{ClientCapabilities, MessageType, ShowMessageParams};
 use scarb_proc_macro_server_types::jsonrpc::RpcResponse;
 use scarb_proc_macro_server_types::methods::ProcMacroResult;
 use smol_str::ToSmolStr;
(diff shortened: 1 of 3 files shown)


src/lang/proc_macros/controller.rs
```diff
@@ -184,6 +193,14 @@ impl ProcMacroClientController {
                 }
 
                 self.set_proc_macro_server_status(db, ServerStatus::Ready(client));
+
+                if client_capabilities.workspace_semantic_tokens_refresh_support() {
```
mkaput, line 197: 1. please add a comment explaining are we doing this here
2. can we make a little indirection here? put this `if` in a `on_need_refresh` method in semantics logic code and just call that method from here. I think this will declutter this `on_response` function

mkaput (changes requested):

src/lang/proc_macros/controller.rs
piotmag769: Did it a bit differently (I don't want to put requesting logic in `ide`, we usually don't do that)

mkaput (approved):

src/lang/proc_macros/controller.rs
```diff
@@ -203,6 +217,22 @@ impl ProcMacroClientController {
         }
     }
 
+    /// Sends `workspace/semanticTokens/refresh` if supported by the client to make sure macros
+    /// declared by proc macros are properly colored.
+    ///
+    /// Usage: should be called when the set of known macros is changed and all plugins with known
+    /// macros are already in the db.
```
Draggu, line 224:
```suggestion
    /// Usage: should be called when the set of known macros is changed and all plugins with known
    /// macros are already in the db.
```
Can be called before as long as mutable `db` handler is not released

src/lang/proc_macros/controller.rs
```diff
@@ -203,6 +217,22 @@ impl ProcMacroClientController {
...
+    fn on_all_known_macros_in_db(
```
Draggu, line 225:
```suggestion
    fn on_supported_macros_response(
```
Will work too when dynamic loading is added.

Draggu (approved):

PR #371, opened 2025-02-26
Add tests for `snforge` macros

integraledelebesgue:
**Stack**:
- #403
- #371 ⬅

### DIFF
--- a/tests/e2e/macros/fixtures.rs
+++ b/tests/e2e/macros/fixtures.rs
@@ -6,6 +6,8 @@ use super::{MacroTest, SCARB_TEST_MACROS_PACKAGE};
 
 pub struct ProjectWithCustomMacros;
 pub struct ProjectWithMultipleCrates;
+pub struct ProjectWithSnforgeUnitTest;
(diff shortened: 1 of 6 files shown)


mkaput:
I actually just started wondering why are we doing these tests... They are going to require regenerating on each snforge update, or we will just forget to do these updates. Is there any benefit of testing this?

integraledelebesgue:
> I actually just started wondering why are we doing these tests... They are going to require regenerating on each snforge update, or we will just forget to do these updates. Is there any benefit of testing this?

I decided to add them because `snforge` macros are the main use case for our proc macro features. We simply need to bump it periodically.

Arcticae (approved):

Arcticae:
@mkaput it would help here if we had something to assess if something "looks like cairo", and for other UI tests as well (cc @piotmag769).

Any ideas? :P

mkaput:
"looks like Cairo" doesn't seem to be a constructive metric of valid execution

Draggu (approved):

PR #376, opened 2025-02-26
Support rename for modules

piotmag769:
All tests after that

### DIFF
--- a/src/ide/navigation/rename.rs
+++ b/src/ide/navigation/rename.rs
@@ -1,16 +1,31 @@
-use std::collections::HashMap;
-
-use anyhow::anyhow;
-use itertools::Itertools;
-use lsp_server::ErrorCode;
-use lsp_types::{Location, RenameParams, TextEdit, WorkspaceEdit};
-
 use crate::lang::db::{AnalysisDatabase, LsSyntaxGroup};
 use crate::lang::defs::SymbolDef;
 use crate::lang::lsp::{LsProtoGroup, ToCairo};
+use crate::lsp::capabilities::client::ClientCapabilitiesExt;
 use crate::lsp::result::{LSPError, LSPResult};
+use std::collections::HashMap;
 
-pub fn rename(params: RenameParams, db: &AnalysisDatabase) -> LSPResult<Option<WorkspaceEdit>> {
+use anyhow::anyhow;
+use cairo_lang_defs::db::DefsGroup;
+use cairo_lang_defs::ids::{ModuleId, SubmoduleId};
+use cairo_lang_filesystem::ids::FileLongId;
+use cairo_lang_syntax::node::TypedStablePtr;
+use cairo_lang_syntax::node::ast::MaybeModuleBody;
+use cairo_lang_utils::LookupIntern;
+use itertools::Itertools;
+use lsp_server::ErrorCode;
+use lsp_types::{
+    ClientCapabilities, DocumentChangeOperation, DocumentChanges, Location, OneOf,
+    OptionalVersionedTextDocumentIdentifier, RenameFile, RenameParams, ResourceOp,
+    TextDocumentEdit, TextEdit, Url, WorkspaceEdit,
+};
+
+// TODO(#381): handle crates separately (files need to be renamed too).
+pub fn rename(
+    params: RenameParams,
+    db: &AnalysisDatabase,
+    client_capabilities: &ClientCapabilities,
+) -> LSPResult<Option<WorkspaceEdit>> {
     let new_name = params.new_name;
 
     // Copied from https://github.com/starkware-libs/cairo/blob/cefadd5ea60d9f0790d71ae14c97d48aa93bd5bf/crates/cairo-lang-parser/src/lexer.rs#L197.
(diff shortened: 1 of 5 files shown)


src/ide/navigation/rename.rs
```diff
@@ -1,16 +1,31 @@
...
-use crate::lang::lsp::{LsProtoGroup, ToCairo};
-use crate::lsp::result::{LSPError, LSPResult};
+use lsp_types::{
+    ClientCapabilities, DocumentChangeOperation, DocumentChanges, Location, OneOf,
+    OptionalVersionedTextDocumentIdentifier, RenameFile, RenameParams, ResourceOp,
+    TextDocumentEdit, TextEdit, Url, WorkspaceEdit,
+};
 
-pub fn rename(params: RenameParams, db: &AnalysisDatabase) -> LSPResult<Option<WorkspaceEdit>> {
+// TODO: handle crates separately (files need to be renamed too).
```
mkaput, line 23: issue?

src/ide/navigation/rename.rs
```diff
@@ -56,9 +63,118 @@ pub fn rename(params: RenameParams, db: &AnalysisDatabase) -> LSPResult<Option<W
...
+    db: &AnalysisDatabase,
+    module_id: ModuleId,
+    client_capabilities: &ClientCapabilities,
+    new_name: &str,
+) -> LSPResult<Option<ResourceOp>> {
+    if !client_capabilities.workspace_edit_rename_resource_support() {
+        return Err(LSPError::new(
+            anyhow!(
+                "Renaming this module requires renaming files, but the editor does \
+                                not support it"
```
mkaput, line 122:
```suggestion
                "Renaming this module requires renaming files, \
                 which is not supported by this editor."
```

src/ide/navigation/rename.rs
```diff
@@ -56,9 +63,118 @@ pub fn rename(params: RenameParams, db: &AnalysisDatabase) -> LSPResult<Option<W
...
+    })?;
+
+    let path = match file_id.lookup_intern(db) {
+        FileLongId::OnDisk(path) => path,
+        FileLongId::Virtual(_) | FileLongId::External(_) => {
+            return Err(LSPError::new(
+                anyhow!("Rename for virtual files are not supported"),
+                ErrorCode::RequestFailed,
+            ));
+        }
```
mkaput, line 139: hmm

we can just ignore renaming the virtual file here and it will be regenerated by the compiler with the new name afterwards, isn't it?

src/ide/navigation/rename.rs
```diff
@@ -56,9 +63,118 @@ pub fn rename(params: RenameParams, db: &AnalysisDatabase) -> LSPResult<Option<W
...
+    };
+
+    let ops = if path.exists() {
+        let old_uri = Url::from_file_path(&path).unwrap();
+        let new_uri =
+            Url::from_file_path(path.with_file_name(format!("{new_name}.cairo"))).unwrap();
```
mkaput, line 145: this may be silly, but let's maybe throw an error telling that we cannot rename remote files? instead of panicking

src/lang/defs/module.rs
```diff
@@ -10,7 +10,7 @@ use crate::lang::db::AnalysisDatabase;
 /// Information about the definition of a module.
 #[derive(Eq, PartialEq)]
 pub struct ModuleDef {
-    id: ModuleId,
+    pub id: ModuleId,
```
mkaput, line 13: to be consistent, avoid doing `pub` here and write a getter `pub fn module_id(&self) -> ModuleId`

mkaput (changes requested):

src/ide/navigation/rename.rs
piotmag769: You will never panic here thought if you had `OnDisk` variant I think. Can do that though

src/ide/navigation/rename.rs
piotmag769: Won't they be regenerated with old names? I don't see why should we support renames in generated files anyways

src/ide/navigation/rename.rs
mkaput: ah right, overlooked the broader context. ignore this comment

src/ide/navigation/rename.rs
mkaput: hmm, but you're renaming module declaration outside this function, aren't you?

src/ide/navigation/rename.rs
piotmag769: #381 will do after project model changes, will be easier to do. Probably will incorporate `ManifestRegistry` into project model too so everything is in one place

Draggu (approved):

src/ide/navigation/rename.rs
```diff
@@ -56,9 +63,118 @@ pub fn rename(params: RenameParams, db: &AnalysisDatabase) -> LSPResult<Option<W
...
+
+    let file_id = db.module_main_file(module_id).map_err(|_| {
+        LSPError::new(anyhow!("Module main file not found"), ErrorCode::RequestFailed)
+    })?;
+
+    let path = match file_id.lookup_intern(db) {
+        FileLongId::OnDisk(path) => path,
+        FileLongId::Virtual(_) | FileLongId::External(_) => {
+            return Err(LSPError::new(
+                anyhow!("Rename for virtual files are not supported"),
```
mkaput, line 136:
```suggestion
                anyhow!("Cannot rename virtual files"),
```

mkaput (approved):

PR #378, opened 2025-02-27
Fix All imports

### LINKED ISSUE #229: Add "fix all imports" quick fix
### Problem

Like in other IDEs we should support fixing all of the import problems and suggesting many options in case of ambiguity in one of the places

### Proposed Solution

_No response_

### Notes

_No response_

Draggu:
close #229

### DIFF
--- a/src/ide/code_actions/mod.rs
+++ b/src/ide/code_actions/mod.rs
@@ -1,7 +1,7 @@
 use cairo_lang_syntax::node::SyntaxNode;
 use lsp_types::{
-    CodeAction, CodeActionOrCommand, CodeActionParams, CodeActionResponse, Diagnostic,
-    NumberOrString, Range, Url,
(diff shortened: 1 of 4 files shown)


src/ide/code_actions/mod.rs
```diff
@@ -50,7 +51,9 @@ fn get_code_actions_for_diagnostics(
 ) -> Vec<CodeAction> {
     let uri = &params.text_document.uri;
 
-    params
+    let mut actions_for_fix_all = vec![];
```
piotmag769, line 54: Add missing trait here

Arcticae (approved):

piotmag769 (approved):
So you decided to collect all?

Draggu:
> So you decided to collect all?

Now it is based on `is_preferred`, added where necessary in #390

PR #389, opened 2025-02-28
Improve the plugin type recognition

integraledelebesgue:
## Changes
* Replaced the hacked logic of discovering the concrete types of plugins given as trait objects with more robust solution


piotmag769 (approved):
❤️

piotmag769 (commented):
Link the issue pls and mentioned related cairo PR in the description

PR #370, opened 2025-02-26
Add tests for procedural macros

integraledelebesgue:
**Stack**:
- #403
- #371
- #370 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #403, opened 2025-03-03
Add tests for builtin macros

integraledelebesgue:
**Stack**:
- #371
- #370
- #402


tests/e2e/macros/builtin/snapshots/e2e__macros__builtin__starknet_interface.snap
```diff
@@ -0,0 +1,769 @@
+---
```
piotmag769, line 1: We shouldn't test starknet macros, they are too unstable. No one will review the snap diffs tbh, so it's like we test nothing imo

tests/e2e/macros/builtin/snapshots/e2e__macros__builtin__attribute_with_inline.snap
```diff
@@ -0,0 +1,68 @@
+---
+source: tests/e2e/macros/builtin/mod.rs
+description: "// → test_package/cairo_project.toml\n[crate_roots]\ntest_package = \"src\"\n\n[config.global]\nedition = \"2024_07\"\n\n// → test_package/src/lib.cairo\n#[derive(Drop)]\nstruct A {\n    a: felt252\n}\n\n#[generate_trait]\nimpl ImplA of ATrait {\n    fn c(self: A) {\n        println!(\"a\");\n        self.a + 1;\n    }\n}"
```
piotmag769, line 3: Can we somehow prettify it?

tests/e2e/macros/builtin/snapshots/e2e__macros__builtin__inline.snap
```diff
@@ -0,0 +1,70 @@
...
+    a
+}
+"""
+
+[[expansions]]
+analyzed_lines = '''
+   <caret>  println!("some text");
+     pr<caret>intln!("some text");
+     println!<caret>("some text");
+'''
```
piotmag769, line 51: What happened here (and in other places as well)?

piotmag769 (changes requested):

Arcticae (changes requested):
https://github.com/software-mansion/cairols/pull/403#discussion_r1983228311
I think the point is to not test expansion logic (where you put the caret) but rather the output (did it return the expected thing) - so we don't need multiple carets in one case (which makes the test less readable and run for longer)

Pls apply to all in the stack

tests/e2e/macros/builtin/snapshots/e2e__macros__builtin__inline.snap
integraledelebesgue: Discussed offline, closing

piotmag769 (approved):

integraledelebesgue:
> [#403 (comment)](https://github.com/software-mansion/cairols/pull/403#discussion_r1983228311) I think the point is to not test expansion logic (where you put the caret) but rather the output (did it return the expected thing) - so we don't need multiple carets in one case (which makes the test less readable and run for longer)
> 
> Pls apply to all in the stack

As we discussed offline, `macros` module contains complex tests of proc macro support, and thus tests with multiple carets are absolutely advisable.

Arcticae (approved):

PR #408, opened 2025-03-04
Find definitions on builtin types

### LINKED ISSUE #306: Find definitions on `usize`, `u128` etc. panics
Same with as it is based on the same mechanism

Draggu:
close #306

### DIFF
--- a/tests/e2e/find_references/mod.rs
+++ b/tests/e2e/find_references/mod.rs
@@ -89,7 +89,10 @@ fn find_references(cairo_code: &str) -> String {
 
             let mut result = String::new();
 
-            if remove_core_references(&mut declarations) || remove_core_references(&mut usages) {
+            let removed_via_declarations = remove_core_references(&mut declarations);
+            let removed_via_references = remove_core_references(&mut usages);
+
+            if removed_via_declarations || removed_via_references {
                 result.push_str("// found several references in the core crate\n");
             }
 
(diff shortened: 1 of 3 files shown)


piotmag769 (approved):
🚀 🔥 😮‍💨 🥇 💯

tests/e2e/find_references/mod.rs
```diff
@@ -89,7 +90,10 @@ fn find_references(cairo_code: &str) -> String {
 
             let mut result = String::new();
 
-            if remove_core_references(&mut declarations) || remove_core_references(&mut usages) {
+            if [remove_core_references(&mut declarations), remove_core_references(&mut usages)]
+                .into_iter()
+                .any(identity)
+            {
```
mkaput, line 96: revert

mkaput (changes requested):

tests/e2e/find_references/mod.rs
Draggu: Tests will fail if we keep old (broken) logic, previously if we found anything in `declarations` we didn't check `usages` at all

mkaput:
```suggestion
            let removed_via_declarations = remove_core_references(&mut declarations);
            let removed_via_references = remove_core_references(&mut usages);
            if removed_via_declarations || removed_via_references {
```

mkaput (approved):

PR #411, opened 2025-03-04
Use the new version of `scarb-proc-macro-server-types` dependency

### LINKED ISSUE #326: Use robust types instead of `String` in `DefinedMacrosResponse` and `WorkspaceProcMacros` in Scarb

integraledelebesgue:
Closes #326 
Requires https://github.com/software-mansion/scarb/pull/2042

## Changes
* `scarb-proc-macro-server-types` has been updated to the temporary version from Git
* LS uses a new type model of the Proc Macro Server which introduces a `Package` type to represent the analyzed packages

### DIFF
--- a/src/lang/proc_macros/plugins/downcast.rs
+++ b/src/lang/proc_macros/plugins/downcast.rs
@@ -41,7 +41,7 @@ mod unsafe_downcast_ref_tests {
     fn cast_succeed() {
         let mut db = AnalysisDatabase::new();
 
-        let context = ProcMacroScope { package_id: String::from("anything") };
+        let context = ProcMacroScope { component: Default::default() };
 
         let input = ExpandAttributeParams {
             context,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2679,8 +2679,7 @@ dependencies = [
 [[package]]
 name = "scarb-proc-macro-server-types"
 version = "0.2.0"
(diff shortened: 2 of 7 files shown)


src/lang/proc_macros/plugins/downcast.rs
```diff
@@ -41,7 +41,8 @@ mod unsafe_downcast_ref_tests {
     fn cast_succeed() {
         let mut db = AnalysisDatabase::new();
 
-        let context = ProcMacroScope { package_id: String::from("anything") };
+        let context =
+            ProcMacroScope { package_id: SerializedPackageId::from("anything".to_string()) };
```
mkaput, line 45:
```suggestion
            ProcMacroScope { package_id: SerializedPackageId::from("anything") };
```

why not?

mkaput (approved):

piotmag769 (commented):
Imo we should break the interface as proposed in the scarb PR

src/lang/proc_macros/plugins/downcast.rs
integraledelebesgue: Changed to `Default::default`, it's enough in this case

piotmag769 (approved):

piotmag769 (commented):
Outdated description

piotmag769 (approved):

PR #419, opened 2025-03-05
Remove generic completions

Draggu:
**Stack**:
- #430
- #429
- #425
- #419 ⬅


integraledelebesgue (approved):

tests/e2e/completions/typing.rs
```diff
@@ -0,0 +1,53 @@
...
+    [[completions]]
+    completion_label = "ByteArrayIter"
+
+    [[completions]]
+    completion_label = "ByteArrayTrait"
+    "#);
+}
+
+#[test]
+fn multi_segment_path() {
```
piotmag769, line 35: This should be in path completions tests

src/ide/completion/expr/typing.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::ide::completion::helpers::completion_kind::importable_completion_kind;
+use crate::lang::analysis_context::AnalysisContext;
+use crate::lang::db::AnalysisDatabase;
+use cairo_lang_semantic::db::SemanticGroup;
+use cairo_lang_syntax::node::TypedSyntaxNode;
+use cairo_lang_syntax::node::ast::ExprPath;
+use cairo_lang_syntax::node::kind::SyntaxKind;
+use if_chain::if_chain;
+
+pub fn typing_completions(db: &AnalysisDatabase, ctx: &AnalysisContext<'_>) -> Vec<CompletionItem> {
```
piotmag769, line 13: Move to path completions

Arcticae, line 13: I think it describes better what's actually happening 

```suggestion
pub fn importable_completions(db: &AnalysisDatabase, ctx: &AnalysisContext<'_>) -> Vec<CompletionItem> {
```

src/ide/completion/expr/typing.rs
```diff
@@ -0,0 +1,46 @@
...
+            return Default::default();
+        }
+    );
+
+    importables
+        .iter()
+        .filter_map(|(importable, path_str)| {
+            let path_segments: Vec<_> = path_str.split("::").collect();
+
+            let last_segment = path_segments.last().expect("path to import should not be empty");
```
Arcticae, line 32: What if user wants to use more than just the last thing? 
i.e. use `std::memcpy` instead of just `memcpy`? 
Can't we suggest the whole thing?

Draggu: #428, Will do later

mkaput (approved):

PR #420, opened 2025-03-05
Rename tests

piotmag769:


tests/e2e/rename/macros.rs
```diff
@@ -0,0 +1,24 @@
...
+    // found renames in the core crate
+    fn main() {
+        print!("Hello world!");
+        let arr = RENAMED![1, 2, 3, 4, 5];
+    }
+    fn bar() {
+        let forty_two = RENAMED![42];
+    }
+    "#)
+}
```
mkaput, line 24: ouch, this isn't something I think we might want to do. you're not renaming the macro itself at all here

mkaput (changes requested):

tests/e2e/rename/macros.rs
piotmag769: I know, but I figured that you may want to do it if you are changing sth in compiler plugins/proc macro plugins

piotmag769: Nvm, it is not in semantics even, will fix it

mkaput (approved):

Draggu (approved):

PR #444, opened 2025-03-07
Improve UX of goto on non-inline modules

### LINKED ISSUE #266: Goto on modules defined in separate file has bad UX
### Problem

When you do goto to a module that is a file, instead of redirecting to a file, it redirects to a name. And then it treats it as a definition, so instead of redirecting you to a file like in rust, it gives you usages.

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Scarb Version

```text

```

piotmag769:
Fixes #266

### DIFF
--- a/src/ide/navigation/goto_definition.rs
+++ b/src/ide/navigation/goto_definition.rs
@@ -1,9 +1,14 @@
-use cairo_lang_utils::Upcast;
-use lsp_types::{GotoDefinitionParams, GotoDefinitionResponse, Location};
-
-use crate::lang::db::{AnalysisDatabase, LsSyntaxGroup};
+use crate::lang::db::{AnalysisDatabase, LsSemanticGroup, LsSyntaxGroup};
 use crate::lang::defs::SymbolDef;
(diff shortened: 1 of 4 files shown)


mkaput (approved):
😍

src/ide/navigation/mod.rs
```diff
@@ -1,4 +1,18 @@
...
+
+use crate::lang::db::AnalysisDatabase;
+
 pub mod goto_definition;
 pub mod highlight;
 pub mod references;
 pub mod rename;
+
+/// Infallible version of `db.is_submodule_inline`.
+fn is_submodule_inline(db: &AnalysisDatabase, submodule_id: SubmoduleId) -> bool {
```
Draggu, line 13: Would fit better in `lang` for potential reuse in completions and code actions

Draggu (approved):

src/ide/navigation/mod.rs
piotmag769: Good idea

PR #465, opened 2025-03-12
Update cairo-vm patch

mkaput:
Now we're targeting upstream.


Cargo.toml
```diff
@@ -127,8 +127,8 @@ cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo", rev = "252
 
 # FIXME(#447): Remove this when https://github.com/lambdaclass/cairo-vm/pull/1988 is shipped.
```
Draggu, line 128: Change comment to remove this when `x.y.z` is released

Draggu (approved):

Cargo.toml
mkaput:
```suggestion
# FIXME(#447): Remove this when cairo-vm 2.0.1 is released.
```

PR #477, opened 2025-03-13
Add code lens tests

Draggu:


mkaput (approved):

mkaput:
I think this might be overthinking and thus not blocking: what about pathological case where somebody writes their own `#[test]` procmacro?

Draggu:
> I think this might be overthinking and thus not blocking: what about pathological case where somebody writes their own `#[test]` procmacro?

As long as there is no proper semantic model for macros, there is nothing we can do with it. Also `snforge/cairo_test` will fail with conflict during `scarb build` anyway.

integraledelebesgue (approved):

PR #501, opened 2025-03-21
Make path_suffix_completions() work with multi segment path

### LINKED ISSUE #428: Make `path_suffix_completions()` work with multi segment path
We should check not only last, but whole type path

Draggu:
Close #428

Also restrict `path_prefix_completions()` to `Self` and paths in `use` statement as it makes no sense now to use this logic in other paths.

### DIFF
--- a/src/ide/completion/path.rs
+++ b/src/ide/completion/path.rs
@@ -1,11 +1,9 @@
-use std::ops::Not;
-
 use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_defs::ids::{FileIndex, ModuleFileId, NamedLanguageElementId};
 use cairo_lang_semantic::db::SemanticGroup;
 use cairo_lang_semantic::diagnostic::{NotFoundItemType, SemanticDiagnostics};
 use cairo_lang_semantic::items::visibility::peek_visible_in;
-use cairo_lang_semantic::resolve::{AsSegments, ResolvedConcreteItem, ResolvedGenericItem};
+use cairo_lang_semantic::resolve::{ResolvedConcreteItem, ResolvedGenericItem};
 use cairo_lang_semantic::{ConcreteTypeId, TypeLongId};
 use cairo_lang_syntax::node::TypedSyntaxNode;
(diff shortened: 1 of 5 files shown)


src/ide/completion/self_completions.rs
```diff
@@ -0,0 +1,28 @@
...
+use lsp_types::CompletionItem;
+
+use if_chain::if_chain;
+
+pub fn self_completions(db: &AnalysisDatabase, ctx: &AnalysisContext<'_>) -> Vec<CompletionItem> {
+    if_chain!(
+        if let Some(expr) = expr_selector(db, &ctx.node);
+        let mut segments = expr.to_segments(db);
+        let _ = {
+            if expr.has_tail(db) {
```
piotmag769, line 14: Can you comment why we do that?

tests/e2e/completions/path.rs
```diff
@@ -79,3 +79,29 @@ fn multi_segment_path() {
...
+    test_transform!(test_completions_text_edits,"
+    mod foo {
+        pub mod bar {
+            pub struct Baz {}
+        }
+    }
+
+    fn a() {
+        bar::B<caret>
+    }
```
piotmag769, line 94:
```suggestion
    mod foo {
        pub mod bar {
            pub struct Baz {}
        }
        pub struct Boo {}
    }

    fn a() {
        bar::B<caret>
    }
```
just to check if it is not catched

src/ide/completion/path.rs
```diff
@@ -54,29 +29,50 @@ pub fn path_suffix_completions(
...
+            let mut last_poped = None;
+
+            let previous_segment_matches = typed_text.iter().rev().all(|typed_segment| {
+                last_poped = path_segments.pop();
+
+                Some(typed_segment.as_str()) == last_poped
+            });
+
+            if !path_segments.is_empty() {
+                path_segments.push(last_poped.unwrap_or(last_segment));
```
piotmag769, line 66: Can we add a note why?

piotmag769 (approved):

integraledelebesgue (approved):

PR #525, opened 2025-03-28
Rename on module file name change

### LINKED ISSUE #427: Rename on module file name change
### Problem

When user is renaming a `*.cairo` file, let's edit accordingly the matching `mod XYZ;` item.

Should be done when receiving `textDocument/didChange`

### Proposed Solution

_No response_

### Notes

_No response_

Draggu:
Close #427

### DIFF
--- a/src/lang/rename_file.rs
+++ b/src/lang/rename_file.rs
@@ -1,0 +1,105 @@
+use super::{db::AnalysisDatabase, lsp::LsProtoGroup};
+use crate::{
+    lang::{defs::SymbolDef, lsp::ToLsp},
+    server::is_cairo_file_path,
+};
+use cairo_lang_defs::{
+    db::DefsGroup,
+    ids::{ModuleId, NamedLanguageElementId},
+};
+use cairo_lang_filesystem::{
+    db::{CrateConfiguration, FilesGroup},
+    ids::Directory,
+};
+use lsp_types::{FileRename, RenameFilesParams, TextEdit, Url, WorkspaceEdit};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
+use tracing::error;
+
+pub fn rename_files(db: &AnalysisDatabase, params: RenameFilesParams) -> Option<WorkspaceEdit> {
+    let mut changes: HashMap<Url, Vec<TextEdit>> = Default::default();
+
+    for rename in params.files {
+        handle_rename(db, rename, &mut changes);
+    }
+
+    Some(WorkspaceEdit { changes: Some(changes), ..Default::default() })
+}
+
+fn handle_rename(
+    db: &AnalysisDatabase,
+    rename: FileRename,
+    changes: &mut HashMap<Url, Vec<TextEdit>>,
+) -> Option<()> {
+    let old_uri = Url::parse(&rename.old_uri).ok()?;
+    let new_uri = Url::parse(&rename.new_uri).ok()?;
+
+    if !is_cairo_file_path(&old_uri) || !is_cairo_file_path(&new_uri) {
+        return None;
+    }
+
+    let file = db.file_for_url(&old_uri)?;
+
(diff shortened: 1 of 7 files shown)


src/lang/rename_file.rs
```diff
@@ -0,0 +1,113 @@
...
+    changes: &mut HashMap<Url, Vec<TextEdit>>,
+) -> Option<()> {
+    let old_uri = Url::parse(&rename.old_uri).ok()?;
+    let new_uri = Url::parse(&rename.new_uri).ok()?;
+
+    if !is_cairo_file_path(&old_uri) || !is_cairo_file_path(&new_uri) {
+        return None;
+    }
+
+    let file = db.file_for_url(&old_uri)?;
```
Arcticae, line 43: is it guaranteed that it's in db on rename?

src/lang/rename_file.rs
```diff
@@ -0,0 +1,113 @@
...
+            prefix.push(comp1);
+        } else {
+            break;
+        }
+    }
+
+    let old_file_name = assert_single_component(old_path.strip_prefix(&prefix).unwrap())?;
+    let new_file_name = assert_single_component(new_path.strip_prefix(&prefix).unwrap())?;
+
+    let mod_name = db
```
Arcticae, line 82: I don't understand, is it not just the file name? 
(and if yes, why bother with this calculation and the assertion below)

Draggu: >I don't understand, is it not just the file name?

it is file path, not just name

>why bother with this calculation and the assertion below

I placed it here because in the future if we support moves (same request) this value will be used, It can be removed now

src/lang/rename_file.rs
Draggu: No, and if it is not we want to skip it as it is not tracked by us

Arcticae (approved):

src/lang/rename_file.rs
```diff
@@ -0,0 +1,113 @@
...
+        .module_submodules(submodule.parent_module(db))
+        .unwrap()
+        .get(&submodule)
+        .unwrap()
+        .name(db);
```
piotmag769, line 87:
```suggestion
    let mod_name = submodule.name_identifier(db);
```
xD

src/lang/rename_file.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    assert_eq!(old_file_name, mod_name.token(db).text(db));
+
+    for usage in
+        SymbolDef::find(db, &mod_name).unwrap().usages(db).include_declaration(true).collect()
```
piotmag769, line 92: Cannot we use `?` for safety instead?

piotmag769 (commented):
The asserts are dangerous imo. We can end up just like with `SymbolDef` xD

src/lang/rename_file.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    Some(WorkspaceEdit { changes: Some(changes), ..Default::default() })
+}
+
+fn handle_rename(
+    db: &AnalysisDatabase,
+    rename: FileRename,
+    changes: &mut HashMap<Url, Vec<TextEdit>>,
+) -> Option<()> {
```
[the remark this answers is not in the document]
Draggu: I can not, this function produces multiple edits for multiple files, so it is not worth changing it

src/lang/rename_file.rs
```diff
@@ -0,0 +1,105 @@
...
+        db.crate_config(first.owning_crate(db))?
+    else {
+        error!("got unexpected virtual directory while renaming files");
+        return None;
+    };
+
+    let old_path = old_uri.to_file_path().ok()?;
+    let new_path = new_uri.to_file_path().ok()?;
+
+    assert!(old_path.starts_with(&root));
```
piotmag769, line 65: useless imo

piotmag769 (approved):

PR #526, opened 2025-03-28
Don't traverse submodules for code lens on modules

piotmag769:
Imo it's not desired if you have submodule with tests deeper in the module tree. That's also how intelijj does it


piotmag769:
Apparently RA does it differently xD

PR #545, opened 2025-04-03
Don't set cfg(test) for corelib from Scarb cache

piotmag769:


src/project/crate_data.rs
```diff
@@ -117,47 +116,6 @@ impl Crate {
         db.set_override_crate_plugins_from_suite(crate_id, interned_plugins);
     }
 
-    /// Construct a [`Crate`] from data already applied to the [`AnalysisDatabase`].
-    ///
-    /// Returns `None` if the crate is virtual or the crate configuration is missing.
-    pub fn reconstruct(db: &AnalysisDatabase, crate_id: CrateId) -> Option<Self> {
```
piotmag769, line 123: Useless now (it was useless before too)

src/project/mod.rs
```diff
@@ -127,15 +130,30 @@ impl ProjectController {
...
+
+                        let mut core_settings = db.crate_config(core_id).unwrap();
+                        core_settings.settings.cfg_set = Some(
+                            core_settings
+                                .settings
+                                .cfg_set
+                                .unwrap_or_default()
+                                .union(&AnalysisDatabase::initial_cfg_set_for_deps()),
+                        );
+                        db.set_crate_config(core_id, Some(core_settings));
```
Draggu, line 148:
```suggestion
                        let maybe_core_settings = db.crate_config(core_id).map(|mut core_settings| {
                            core_settings.settings.cfg_set = Some(
                                core_settings
                                    .settings
                                    .cfg_set
                                    .unwrap_or_default()
                                    .union(&AnalysisDatabase::initial_cfg_set_for_deps()),
                            );
                            
                            core_settings
                        });
                        db.set_crate_config(core_id, maybe_core_settings);
```

src/project/mod.rs
```diff
@@ -298,3 +316,11 @@ struct ProjectUpdateRequest {
     file_path: PathBuf,
     loaded_manifests: Snapshot<HashSet<PathBuf>>,
 }
+
+fn contains_core_from_scarb_cache(
+    project_config: &ProjectConfig,
+    scarb_toolchain: &ScarbToolchain,
+) -> bool {
+    project_config.content.crate_roots.keys().contains(&CrateIdentifier::from("core"))
```
Arcticae, line 324: Isn't it possible to check the core crate's location directly? 
It seems hacky to check `project_config.base_path` like this 🤔 
Is it even correct?

src/project/mod.rs
```diff
@@ -127,15 +130,30 @@ impl ProjectController {
                 );
             }
             ProjectUpdate::CairoProjectToml(maybe_project_config) => {
+                if let Some(project_config) = maybe_project_config {
```
Arcticae, line 133: Question: Why do we only add this logic on the `CairoProjectToml` update, when corelib has both this and `Scarb.toml` ? Is it guaranteed to have precedence?

Arcticae (approved):

integraledelebesgue (approved):

src/project/mod.rs
piotmag769: I guess you are right, I will use the same logic that is used [here](https://github.com/starkware-libs/cairo/blob/e1ae8f2744eeedc54dbd96924c0ac73479e98c8b/crates/cairo-lang-project/src/lib.rs#L75)

src/project/mod.rs
piotmag769: I can unwrap since the `ProjectConfig` was already used to apply changes to the db and I checked that it contains corelib. And if `maybe_core_settings` will ever be `None` (won't be) it will remove the crate from `crate_configs`. Imo this is more confusing than my previous code

src/project/mod.rs
```diff
@@ -127,15 +130,30 @@ impl ProjectController {
...
+                    // Make sure cfg(test) is not set if the core crate comes from the Scarb cache.
+                    if contains_core_from_scarb_cache(&project_config, &state.scarb_toolchain) {
+                        let core_id = CrateId::core(db);
+
+                        let mut core_settings = db.crate_config(core_id).unwrap();
+                        core_settings.settings.cfg_set = Some(
+                            core_settings
+                                .settings
+                                .cfg_set
+                                .unwrap_or_default()
```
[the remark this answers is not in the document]
piotmag769: >Shouldn't this initial_cfg_set_for_deps thing be the default?

No, because `cairo_project.toml` projects would lose the ability to analyse code tagged with `#[cfg(test)]`

>Or at least aliased somehow

wdym?

Draggu (approved):

PR #551, opened 2025-04-05
Fix code lens snforge command for modules

piotmag769:


src/ide/code_lens/tests.rs
```diff
@@ -167,16 +188,12 @@ impl TestRunner {
...
             },
-            Self::CairoTest if available_runners.cairo_test => {
-                Some(format!("scarb cairo-test --filter {test_path}"))
-            }
-            Self::Snforge if available_runners.snforge => {
-                Some(format!("snforge test {test_path} --exact"))
-            }
+            Self::CairoTest if available_runners.cairo_test => Some(test_path.cairo_test_command()),
+            Self::Snforge if available_runners.snforge => Some(test_path.snforge_command()),
             Self::Custom => Some(custom_command.replace("{{TEST_PATH}}", test_path.as_ref())),
```
Draggu, line 197: If there is a need for commands to know if this is module or function we should provide this data to custom command as well. Form of this can be discussed and implemented in another PR though.

Draggu (approved):

src/ide/code_lens/tests.rs
Arcticae: What's the use case right now though?

Arcticae (approved):

src/ide/code_lens/tests.rs
piotmag769: So we know when to pass `--exact`, but it's not a big deal

Draggu: If you want to add extra args to snforge you have to use custom command, there is a need to know if it was module or function so you can apply `--exact` which is making compile optimizations

piotmag769:
Waiting for scarb dev so I can fix tests

Issue #170, opened 2025-01-13
Finding references and rename for impl methods does not traverse usages via trait


mkaput:
Tagged with `FIXME(#170)`

```cairo
#[derive(Drop)]
struct Foo {}

trait FooTrait {
    fn area(self: @Foo) -> u64;
}

impl FooImpl of FooTrait {
    fn are<caret>a(self: @Foo) -> u64 { 0 }
}

fn main() {
    let foo = Foo {};
    let x = foo.area();
}
```

This does not return `foo.<sel>area</sel>()`. It also should pick up `area` in `FooTrait` definition (to be discussed - would also affect goto).

Note that the expected outcome is to find all references to **this implementation method**. This means that if `FooTrait` would also be implemented for another struct, `Bar`, the usage `bar.area()` would not be picked up because it's not routed through `FooImpl`.

The same applies for alternative implementations of a same trait for same struct (say, `FooImplBis`).

Also when caret is on 

```cairo
trait FooTrait {
    fn a<caret>rea(self: @Foo) -> u64;
}
```

area inside `FooImpl` should be detected as definition/usage as well in this context.

mkaput:
1. Goto definition: go from impl def to trait def
2. Find references:
    1. on impl def: find concrete references, or if user wants to, they can goto trait def and from that find all references
    2. on usage: find concrete references
4. Rename: rename all

mkaput:
The same applies to associated items such as consts

PR #570, opened 2025-04-09
Support goto definition for trait impl items

Arcticae:
Related: [#170](https://github.com/software-mansion/cairols/issues/170)


tests/e2e/goto_definition/consts.rs
```diff
@@ -39,21 +39,6 @@ fn associated_const_via_trait_declaration() {
...
-    struct Triangle {}
-    impl TriangleShape of Shape<Triangle> { const SIDE<caret>S: u32 = 3; }
-    "#, @r"
-    trait Shape<T> { const SIDES: u32; }
-    struct Triangle {}
-    impl TriangleShape of Shape<Triangle> { const <sel>SIDES</sel>: u32 = 3; }
-    ");
-}
-
-// FIXME: https://github.com/software-mansion/cairols/issues/405
```
Arcticae, line 55: This turns out that it was fine overall xD

tests/e2e/goto_definition/consts.rs
```diff
@@ -39,21 +39,6 @@ fn associated_const_via_trait_declaration() {
     "#, @"trait Shape<T> { const <sel>SIDES</sel>: u32; }");
 }
 
-#[test]
-fn associated_const_via_impl_definition() {
```
Arcticae, line 43: This was moved to `impl.rs`

tests/e2e/goto_definition/traits.rs
```diff
@@ -53,27 +53,6 @@ fn full_path_trait_name_in_expr() {
     ")
 }
 
-#[test]
-fn method_in_impl() {
```
Arcticae, line 57: moved to `impls.rs`

src/lang/defs/finder.rs
```diff
@@ -47,10 +47,66 @@ pub fn find_definition(
...
+        .intern(db);
+        let trait_id = db.impl_def_concrete_trait(long_id).ok()?.trait_id(db);
+
+        if let Some(function) =
+            identifier.as_syntax_node().ancestor_of_type::<ast::FunctionDeclaration>(db)
+        {
+            // We don't want this to look for the trait definition on parameters themselves
+            if identifier.as_syntax_node().ancestor_of_type::<ast::ParamList>(db).is_some() {
+                return None;
+            }
```
Arcticae, line 74: idk if this is acceptable but i think it's what we want?

integraledelebesgue: Is this a situation like:
```cairo
impl [...] {
    fn foo(x<caret>: felt252) {}
}
```
?

If yes, we should indeed return `None`. Generic parameters are the analogical case, we want to skip them too. I think we could simply check if the `ast::FunctionDeclaration` is the **direct** parent of the identifier instead of using `ancestor_of_kind`

Draggu (approved):

src/lang/defs/finder.rs
```diff
@@ -47,10 +47,66 @@ pub fn find_definition(
...
         .or_else(|| lookup_resolved_items(db, identifier, lookup_items))
         .or_else(|| lookup_item_name(db, identifier, lookup_items))
 }
 
+/// Resolve elements of `impl`s to trait definitions.
+fn try_impl_items(
+    db: &AnalysisDatabase,
+    identifier: &ast::TerminalIdentifier,
+) -> Option<ResolvedItem> {
+    if let Some(item_impl) = &identifier.as_syntax_node().ancestor_of_type::<ast::ItemImpl>(db) {
```
integraledelebesgue, line 60: I think it would be more readable since it has one lower level of nesting:
```suggestion
    let Some(item_impl) = &identifier.as_syntax_node().ancestor_of_type::<ast::ItemImpl>(db) else {
        return None;
    }
    
    // [...]  whole function code
```

piotmag769 (approved):

piotmag769 (approved):

PR #571, opened 2025-04-10
Test LSP features for extern types and type aliases

### LINKED ISSUE #438: Add extensive hover/goto/find references/rename tests for standalone types

integraledelebesgue:
Closes #438

### DIFF
--- a/tests/e2e/find_references/types.rs
+++ b/tests/e2e/find_references/types.rs
@@ -1,0 +1,717 @@
+use crate::find_references::find_references;
+use crate::support::insta::test_transform;
+
+#[test]
+fn test_extern_type_in_alias_as_type() {
+    test_transform!(find_references, r#"
+    type TypeAlias = u32<caret>;
+    "#, @r"
+    // found several references in the core crate
+    type TypeAlias = <sel>u32</sel>;
+    ")
+}
+
+#[test]
+fn test_extern_type_in_alias_as_type_parameter() {
+    test_transform!(find_references, r#"
+    type TypeAlias = Array<u32<caret>>;
+    "#, @r"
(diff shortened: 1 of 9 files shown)


generate_test.py
```diff
@@ -0,0 +1,202 @@
+from textwrap import dedent, indent
```
integraledelebesgue, line 1: I will delete it later, before merging

Draggu:
Broken formatting in test code strings

Draggu (approved):

tests/e2e/find_references/types.rs
```diff
@@ -0,0 +1,717 @@
...
+        type TypeAlias = Struct;
+    }
+    use mod1::TypeAlias<caret>;
+    "#, @r"
+    mod mod1 {
+        struct Struct {}
+        type <sel=declaration>TypeAlias</sel> = Struct;
+    }
+    use mod1::<sel>TypeAlias</sel>;
+    ")
```
piotmag769, line 325: Lol, and it somehow works here while other aliases do not xD

piotmag769 (approved):
Codegen good, honestly we can even leave this file in the repo in `codegen/` folder if you wish so

tests/e2e/find_references/types.rs
integraledelebesgue: Indeed - that's because the resolution of paths in `use` is different than everywhere else

integraledelebesgue:
> Codegen good, honestly we can even leave this file in the repo in `codegen/` folder if you wish so

I think it's no longer necessary. Also, I don't want to pollute the codebase with Python
🐍 < 🦀

PR #594, opened 2025-04-18
Fix lost diagnostics from macros

Arcticae:
They were dropped if they were mapping to another file

### DIFF
--- a/src/lang/diagnostics/file_diagnostics.rs
+++ b/src/lang/diagnostics/file_diagnostics.rs
@@ -18,7 +18,6 @@ use tracing::{error, info_span};
 
 use crate::lang::db::AnalysisDatabase;
 use crate::lang::diagnostics::lsp::map_cairo_diagnostics_to_lsp;
-use crate::lang::lsp::LsProtoGroup;
 use crate::server::panic::is_cancelled;
 
 /// Result of refreshing diagnostics for a single file.
@@ -34,25 +33,9 @@ use crate::server::panic::is_cancelled;
 /// to the given `file` will also be visited and their diagnostics collected.
 #[derive(Clone, PartialEq, Eq)]
 pub struct FileDiagnostics {
-    /// The file URL these diagnostics are associated with.
-    pub url: Url,
     pub parser: Diagnostics<ParserDiagnostic>,
     pub semantic: Diagnostics<SemanticDiagnostic>,
     pub lowering: Diagnostics<LoweringDiagnostic>,
-}
-
-pub enum LSPDiagnostic {
-    Standard(Diagnostic),
-    Remapped(Diagnostic),
-}
-
-impl LSPDiagnostic {
-    pub(crate) fn unpack(self) -> Diagnostic {
-        match self {
-            LSPDiagnostic::Standard(x) => x,
-            LSPDiagnostic::Remapped(x) => x,
-        }
-    }
 }
 
 impl FileDiagnostics {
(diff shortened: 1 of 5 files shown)


src/lang/diagnostics/project_diagnostics.rs
```diff
@@ -31,28 +30,90 @@ impl ProjectDiagnostics {
...
+    pub fn apply_updates(
+        &self,
+        updates: HashMap<Url, Vec<LSPDiagnostic>>,
+        scarb_toolchain: &ScarbToolchain,
+    ) {
+        for (file, lsp_diagnostics) in updates {
+            // There are basically ONLY 2 cases:
+            // (if there's something else here something went seriously wrong)
+            // 1. Remapped entry - those we want to append to existing diagnostics
+            // 2. Non-remapped ones - we want to substitute the original ones with them
```
piotmag769, line 61: ?

piotmag769: Nvm, it is half true, but the whole Remapped/Standard is not needed:
1. The comment is inconsistent with the code: `diags.clone_into(&mut new_diags)` clones `diags` (old ones) into `new_diags` (new ones), effectively ignoring new diagnostics.
2. Anyways, in every case we should just substitute the existing diagnostics with new ones: diags for virtual files are always generated from diags from exactly ONE on disk file (besides parser diagnostics, they are generated from the virtual file themselves but those somehow work and I have no idea why).

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -41,6 +41,20 @@ pub struct FileDiagnostics {
...
+
+impl LSPDiagnostic {
+    pub(crate) fn unpack(self) -> Diagnostic {
+        match self {
+            LSPDiagnostic::Standard(x) => x,
+            LSPDiagnostic::Remapped(x) => x,
+        }
+    }
+}
+
```
piotmag769, line 57: Unnecessary as stated in another comment

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -121,8 +128,8 @@ impl FileDiagnostics {
         db: &AnalysisDatabase,
         file_id: FileId,
         trace_macro_diagnostics: bool,
-    ) -> lsp_types::PublishDiagnosticsParams {
-        let mut diagnostics = Vec::new();
+    ) -> HashMap<Url, Vec<LSPDiagnostic>> {
```
piotmag769, line 131: The `HashMap` here breaks the entire diagnostics logic 🙃 It prevents clearing of diagnostics when everything goes right (no key in hashmap - no update). That's the main problem with this PR

piotmag769 (changes requested):

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -96,55 +95,44 @@ impl FileDiagnostics {
             }
         }
 
-        let parser_file_diagnostics = query!(db.file_syntax_diagnostics(file)).unwrap_or_default();
+        // TODO: collect parser diags for virtual files
```
piotmag769, line 98: This never worked xD

src/lang/diagnostics/refresh.rs
```diff
@@ -58,20 +57,34 @@ fn refresh_file_diagnostics(
...
-    let is_dependency = scarb_toolchain.cache_path().is_some_and(|cache_path| {
-        params.uri.to_file_path().is_ok_and(|p| p.starts_with(cache_path))
-    });
-    if is_dependency {
-        params.diagnostics.retain(|diag| diag.severity == Some(DiagnosticSeverity::ERROR));
-    }
+    let new_diags = new_diags
+        .into_iter()
+        .filter_map(|(file, mut diagnostics)| {
+            // TODO: fix for virtual files from deps
```
piotmag769, line 65: Same, never worked, we need to look for parent for virtual files

piotmag769:
Closing in favour of #608 - copied the logic that did in the last commit here

PR #620, opened 2025-04-30
Fix derive macros

piotmag769:


src/lang/proc_macros/db.rs
```diff
@@ -97,14 +97,13 @@ pub fn get_derive_expansion(
     params: ExpandDeriveParams,
 ) -> ProcMacroResult {
     db.get_stored_derive_expansion(params.clone().into()).unwrap_or_else(|| {
-        let token_stream = params.item.clone();
```
piotmag769, line 100: I have no idea how it could ever work in tests btw since this token stream containst the `#[derive(MacroToBeExpanded)]` part

Draggu: Please write an explanation of why it happened (cancellation) that you shared with us.

Draggu (approved):

src/lang/proc_macros/db.rs
```diff
@@ -97,14 +97,13 @@ pub fn get_derive_expansion(
...
-
         if let Some(client) = db.proc_macro_server_status().ready() {
             client.request_derives(params);
         }
 
         ProcMacroResult {
-            token_stream: token_stream_v2_to_v1(&token_stream),
+            // We don't remove the original item for derive macros, so return nothing.
```
Arcticae, line 105: If we don't remove anything, there should be the original item here, why do you return an empty string instead?

Arcticae (approved):

PR #638, opened 2025-05-12
Fix parser and vfs modules diagnostics

### LINKED ISSUE #142: Unnecessary clearing of diagnostics
Currently if there is a panic during diagnostics calculation we log an error and use a default value (empty diagnostics).

https://github.com/software-mansion/cairols/blob/05a22493597b4638137c545e177ac3b80efbc011/src/lang/diagnostics/file_diagnostics.rs#L82-L85

This can cause blinking of diagnostics bcs we may clear already existing ones. Instead we should use previously calculated diagnostics to not clear any existing diagnostics

piotmag769:
Closes #142

---

**Stack**:
- #646
- #645
- #638 ⬅

### DIFF
--- a/src/lang/diagnostics/file_diagnostics.rs
+++ b/src/lang/diagnostics/file_diagnostics.rs
@@ -1,8 +1,5 @@
-use std::collections::{HashMap, HashSet};
-use std::panic::{AssertUnwindSafe, catch_unwind, resume_unwind};
+use std::collections::HashMap;
 
-use cairo_lang_defs::db::DefsGroup;
-use cairo_lang_defs::ids::ModuleId;
 use cairo_lang_diagnostics::Diagnostics;
 use cairo_lang_filesystem::db::FilesGroup;
 use cairo_lang_filesystem::ids::FileId;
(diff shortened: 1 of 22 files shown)


src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -31,85 +27,77 @@ use crate::server::panic::is_cancelled;
...
-                        .map(Result::unwrap_or_default)
-                        .unwrap_or_default()
-                        .get_all(),
-                );
-                lowering_file_diagnostics.extend(
-                    query!(db.module_lowering_diagnostics(module_id))
-                        .map(Result::unwrap_or_default)
-                        .unwrap_or_default()
-                        .get_all(),
-                );
```
piotmag769, line 94: Removed `query!` this since it may lead to temporary diagnostics blinking if this query gets cancelled and previously there were any diagnostics there. This effect was not noticed in practice but this also proves that removing it shouldn't affect anything (if it took noticeable effect then the diagnostics would be blinking)

tests/e2e/macros/procedural/custom/v2/snapshots/e2e__macros__procedural__custom__v2__derive_improper_v2_with_user_error.snap
```diff
@@ -178,3 +266,16 @@ mapped_diagnostics:
               line: 0
               character: 50
         message: Diagnostic mapped from here.
+- url: vfs://proc_macro_derive.cairo
+  diagnostics:
+  - related_code: $>
```
piotmag769, line 271: Bug in scarb or compiler

src/lang/diagnostics/refresh.rs
```diff
@@ -23,51 +22,47 @@ pub fn refresh_diagnostics(
     notifier: Notifier,
     scarb_toolchain: ScarbToolchain,
 ) {
-    let mut processed_modules: HashSet<ModuleId> = HashSet::default();
```
piotmag769, line 26: This is useless now, modules can never repeat if we calculate diagnostics per (on disk file + its descendants)

Draggu (approved):

PR #644, opened 2025-05-13
Fix trait type `SymbolDef`

### LINKED ISSUE #170: Finding references and rename for impl methods does not traverse usages via trait

Arcticae:
Closes #170 

- [x] Take impl impls themselves into account (their usage must be found properly when referred to - just like type in this PR)

### DIFF
--- a/src/lang/db/semantic.rs
+++ b/src/lang/db/semantic.rs
@@ -4,7 +4,7 @@ use cairo_lang_defs::ids::{
     FreeFunctionLongId, ImplAliasLongId, ImplConstantDefLongId, ImplDefLongId, ImplFunctionLongId,
     ImplItemId, LanguageElementId, LookupItemId, ModuleFileId, ModuleId, ModuleItemId,
     ModuleTypeAliasLongId, StructLongId, TraitConstantLongId, TraitFunctionLongId, TraitImplLongId,
-    TraitItemId, TraitLongId, UseLongId, VarId,
+    TraitItemId, TraitLongId, TraitTypeLongId, UseLongId, VarId,
 };
 use cairo_lang_filesystem::ids::FileId;
 use cairo_lang_semantic::Binding;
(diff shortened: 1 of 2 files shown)


Draggu (approved):

src/lang/db/semantic.rs
```diff
@@ -324,6 +324,15 @@ fn lookup_item_from_ast(
...
         }
+        SyntaxKind::TraitItemType => {
+            vec![LookupItemId::TraitItem(TraitItemId::Type(
+                TraitTypeLongId(
+                    module_file_id,
+                    ast::TraitItemType::from_syntax_node(db, node).stable_ptr(db),
+                )
+                .intern(db),
+            ))]
+        }
```
piotmag769, line 335: This only changed fixed stuff in our find ref tests - do we not have tests for associated types for: rename, goto and highlight?

piotmag769 (approved):

tests/e2e/find_references/traits.rs
```diff
@@ -392,6 +406,136 @@ fn dual_implementations_of_trait_via_trait_impl() {
...
+
+    impl GeorgeImpl of BushTrait {
+        impl BushImplementation = GeorgeBush;
+    }
+
+
+    fn main() {
+        let v1 = GeorgeImpl::BushImplemen<caret>tation::bush();
+        let v2 = GeorgeImpl::BushImplementation::bush();
+    }
```
piotmag769, line 440: Add 
```cairo
impl KateImpl of BushTrait {
    impl BushImplementation = KateBush;
}

// main
KateBush::BushImplementation::bush();
```

piotmag769 (approved):
I think we need to review all tests at some point and make sure they cover everything

src/lang/db/semantic.rs
Arcticae: https://github.com/software-mansion/cairols/issues/659

piotmag769 (approved):

integraledelebesgue (approved):

PR #646, opened 2025-05-13
Fix panic when restarting proc macro server due to error

piotmag769:


src/lang/proc_macros/controller.rs
```diff
@@ -171,6 +171,7 @@ impl ProcMacroClientController {
         match db.proc_macro_server_status() {
             ServerStatus::Starting(client) => {
                 let Ok(defined_macros) = client.finish_initialize() else {
+                    drop(client);
```
piotmag769, line 174: These drops fix panic on `Arc::try_unwrap` in this file. When I wrote it I commented correctly that
`At this point we are the only thread with access to the db and therefore to the proc macro client.` but I forgot that another `Arc` with the client may still exist in the caller function.

Draggu (approved):
:rocket:

Arcticae (approved):

PR #650, opened 2025-05-15
Re-add spans to diagnostics calculation

piotmag769:
Useful for profiling


Draggu (approved):

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -39,6 +43,24 @@ impl FilesDiagnostics {
...
+                            error!(
+                                "caught panic when computing diagnostics for file: {:?}",
+                                root_on_disk_file.lookup_intern(db)
+                            );
+                            err
+                        }
+                    })
+                })
+            };
+        }
```
piotmag769, line 62: Re-added it bcs the tracing happened to completely break LS for some reason - I guess it was due to panics in the tracing spans which caused problems. Mind that if a query fails now we don't default to no diagnostics but we simply abort processing this file via `.ok()?` so #142 is still solved

Arcticae (approved):
idk why tests fail tho

PR #664, opened 2025-05-21
Fix tests with broken syntax

Arcticae:


tests/e2e/find_references/types.rs
```diff
@@ -646,6 +696,33 @@ fn test_type_alias_in_impl_associated_type_as_type_parameter() {
     ")
 }
 
+#[test]
+fn test_type_alias_in_impl_associated_type_via_usage() {
```
piotmag769, line 700: Where type alias? I assume you meant
```cairo
trait Trait {
    type Type;
}
impl Impl of Trait {
    type Type = SomeAlias;
}

fn foo() {
    let _v: Impl::Ty<caret>pe = 123;
}
```

piotmag769 (approved):

tests/e2e/find_references/types.rs
Arcticae: yes, corrected. doesn't matter that much for this test what's on rhs though :P

PR #665, opened 2025-05-21
Fix renames for traits 

### LINKED ISSUE #648: Fix behavior of `rename` with impl items
This should work differently than `find references` does - instead of renaming just usages of a specific impl - it should rename the trait as well, as other usages of other impls

(this is most likely what the user needs done, renaming just impl and usages won't make sense during the compilation after the rename - making it effectively invalid)

Arcticae:
closes #648

### DIFF
--- a/src/ide/navigation/rename.rs
+++ b/src/ide/navigation/rename.rs
@@ -41,7 +41,9 @@ pub fn rename(
         let position = params.text_document_position.position.to_cairo();
         let identifier = db.find_identifier_at_position(file, position)?;
 
-        SymbolSearch::find_definition(db, &identifier).map(|search| search.def)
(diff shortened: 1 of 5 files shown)


src/ide/navigation/rename.rs
```diff
@@ -41,7 +41,7 @@ pub fn rename(
         let position = params.text_document_position.position.to_cairo();
         let identifier = db.find_identifier_at_position(file, position)?;
 
-        SymbolSearch::find_definition(db, &identifier).map(|search| search.def)
+        SymbolSearch::find_declaration(db, &identifier).map(|search| search.def)
```
piotmag769, line 44: Please doc why

piotmag769 (approved):

PR #678, opened 2025-05-26
Resolve traits used as bounds for generic type parameters

### LINKED ISSUE #651: Implement LSP features for traits in generic bounds
LSP features (hovers, goto, find references and rename) should work on traits referenced as generic bounds in functions, structs, traits and impls.

## User-defined traits:

Invoking "Go to definition" on:
```cairo
trait GenericTrait<T> {}
fn foo<T, +Generic<caret>Trait<T>>() {}
```
should yield the definition of `GenericTrait`.

## Compiler built-in traits:
```cairo
fn foo<T, +Drop<caret><T>> {}
```
should lead to the definition of `Drop` in `core::traits`.

### LINKED ISSUE #652: Add completions for traits in generic bounds
When typing:
```cairo
fn foo<T, +D<caret>>() {}
```
completion fails and LS panics. The error appears for both built-in and user-defined traits.

Instead, `Drop`, `Destruct`, `Deserialize` etc. should be proposed.

integraledelebesgue:
Closes #651 
Closes #652

### DIFF
--- a/src/lang/defs/finder.rs
+++ b/src/lang/defs/finder.rs
@@ -22,6 +22,7 @@ use cairo_lang_semantic::{ConcreteImplId, Expr, TypeLongId};
 use cairo_lang_syntax::node::ast::TerminalIdentifier;
 use cairo_lang_syntax::node::helpers::{GetIdentifier, HasName};
 use cairo_lang_syntax::node::ids::SyntaxStablePtrId;
+use cairo_lang_syntax::node::kind::SyntaxKind;
 use cairo_lang_syntax::node::{SyntaxNode, Terminal, TypedStablePtr, TypedSyntaxNode, ast};
 use cairo_lang_utils::smol_str::SmolStr;
 use cairo_lang_utils::{Intern, LookupIntern};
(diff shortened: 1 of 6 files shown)


src/lang/defs/finder.rs
```diff
@@ -481,6 +483,51 @@ fn try_concrete_type_or_impl(
...
+            .resolve_generic_path_with_args(
+                &mut SemanticDiagnostics::default(),
+                type_path_segments.clone(),
+                NotFoundItemType::Identifier,
+            )
+            .ok()?;
+
+        match &resolved_item {
+            ResolvedGenericItem::Trait(_) | ResolvedGenericItem::Impl(_) => {
+                return Some(ResolvedItem::Generic(resolved_item));
```
piotmag769, line 522: I wonder if this wouldn't cause weird completions when trait is fully resolved, but I guess it's god enough for now

piotmag769 (approved):

src/lang/defs/finder.rs
Draggu: We should try concrete first, then generic, also `lookup_resolved_generic_item_by_ptr`/`lookup_resolved_concrete_item_by_ptr` should do the job here instead of resolver. (correct me if im wrong)

src/lang/defs/finder.rs
```diff
@@ -481,6 +483,51 @@ fn try_concrete_type_or_impl(
...
+
+    let module_file_id = db.find_module_file_containing_node(&identifier.as_syntax_node())?;
+
+    for &lookup_item_id in lookup_items {
+        let type_path_segments =
+            identifier.as_syntax_node().ancestor_of_type::<ast::ExprPath>(db)?.to_segments(db);
+
+        if type_path_segments
+            .last()
+            .is_none_or(|segment| segment.identifier(db) != identifier.text(db))
```
Draggu, line 504:
```suggestion
        let last_segment = type_path_segments.last()?;
            
        if last_segment.identifier(db) != identifier.text(db)
```

src/lang/defs/finder.rs
integraledelebesgue: Sadly, `lookup_..._item_by_ptr` functions don't work here

PR #694, opened 2025-05-28
Use resultants in semantic highlighting resolver lookups

Arcticae:
This fixes constants and structs not highlighting in v2 macros

### DIFF
--- a/src/ide/semantic_highlighting/token_kind.rs
+++ b/src/ide/semantic_highlighting/token_kind.rs
@@ -256,15 +256,15 @@ fn get_resultants_and_closest_terminals(
         return vec![];
     };
 
-    // First resultant is the best bet to find the correct semantics
     resultants
         .into_iter()
         .filter_map(|resultant| {
             let terminal = if resultant.kind(db).is_terminal() {
                 Some(resultant)
+            } else if resultant.kind(db).is_token() {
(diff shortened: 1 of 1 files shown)


src/ide/semantic_highlighting/token_kind.rs
```diff
@@ -242,3 +249,24 @@ impl SemanticTokenKind {
...
+
+// Retrieves the most-likely-usable resultant, and the terminal ptr we can use for semantic lookup
+fn get_resultant_and_closest_terminal(
+    db: &AnalysisDatabase,
+    node: SyntaxNode,
+) -> Option<(SyntaxNode, TerminalIdentifierPtr)> {
+    let resultants = db.get_node_resultants(node)?;
+
+    // First resultant is the best bet to find the correct semantics
+    let resultant = resultants.first()?;
```
Draggu, line 261: We should first search for semantic, then use first. This way if first resultant does not have semantic (for example macro generated invalid code) we will try next resultant and use first one with semantic meaning

Draggu (changes requested):

src/ide/semantic_highlighting/token_kind.rs
Arcticae: @Draggu pls check if this is what you meant - otherwise idk how to check if an element semantics exist beforehand (which i can do as well here if it makes more sense)

piotmag769 (approved):

PR #700, opened 2025-05-30
Fix AnalysisProgressController

Draggu:
it ignored case where macros where resolved in next generation

### DIFF
--- a/src/lang/proc_macros/client/mod.rs
+++ b/src/lang/proc_macros/client/mod.rs
@@ -216,3 +216,9 @@ impl Iterator for Responses<'_> {
         Some((params, response))
     }
 }
+
+impl Responses<'_> {
+    pub fn len(&self) -> usize {
+        self.responses.len()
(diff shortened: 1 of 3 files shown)


src/lang/proc_macros/client/mod.rs
```diff
@@ -215,4 +215,10 @@ impl Iterator for Responses<'_> {
 
         Some((params, response))
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let size = self.responses.len();
+
+        (size, Some(size))
```
piotmag769, line 222: Just impl `len` on `Reponses` imo

piotmag769 (approved):

src/lang/proc_macros/client/mod.rs
integraledelebesgue: Alternatively, you can write a macro to extract the length

PR #736, opened 2025-06-10
Rename tests in macros

piotmag769:


tests/e2e/support/transform.rs
```diff
@@ -61,11 +62,15 @@ pub fn conduct_transformation<T: Transformer>(
...
 }
 
 pub trait Transformer {
     fn capabilities(base: ClientCapabilities) -> ClientCapabilities;
 
-    fn transform(ls: MockClient, cursors: Cursors) -> String;
+    fn transform(
+        ls: MockClient,
+        cursors: Cursors,
+        additional_data: Option<serde_json::Value>,
```
piotmag769, line 74: It's ugly (kinda untyped) but I don't have a better idea that would be extensible

tests/e2e/support/transform.rs
```diff
@@ -12,6 +12,7 @@ pub fn conduct_transformation<T: Transformer>(
     cairo_code_with_cursors: &str,
     with_macros: bool,
     initial_fixture: Fixture,
+    additional_data: Option<serde_json::Value>,
```
IFoundCursedPRStackThatWillBanYou, line 15:
```suggestion
    config: Option<serde_json::Value>,
```

IFoundCursedPRStackThatWillBanYou (changes requested):

PR #745, opened 2025-06-11
Port adapting call sites from scarb

Arcticae:
<25.06 12:16 - need new scarb-dev for this to work>

### DIFF
--- a/src/lang/proc_macros/db.rs
+++ b/src/lang/proc_macros/db.rs
@@ -80,7 +80,7 @@ pub fn get_attribute_expansion(
     db: &dyn ProcMacroGroup,
     mut params: ExpandAttributeParams,
 ) -> ProcMacroResult {
-    let stabilizer = SpansStabilizer::new(&mut params.call_site, &mut params.item);
+    let stabilizer = SpansStabilizer::new(&mut params.adapted_call_site, &mut params.item);
 
     let result = db.get_stored_attribute_expansion(params.clone().into()).unwrap_or_else(|| {
         let token_stream = params.item.clone();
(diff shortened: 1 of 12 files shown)


src/lang/proc_macros/db.rs
```diff
@@ -78,7 +78,7 @@ pub fn get_attribute_expansion(
     db: &dyn ProcMacroGroup,
     mut params: ExpandAttributeParams,
 ) -> ProcMacroResult {
-    let stabilizer = SpansStabilizer::new(&mut params.call_site, &mut params.item);
+    let stabilizer = SpansStabilizer::new(&mut params.adapted_call_site, &mut params.item);
```
piotmag769, line 81: 1. Why only here? What about derive and inline?
2. Are you sure that only `adapted_call_site` need to be stabilized? If so add a comment why.

src/lang/proc_macros/plugins/scarb/regular.rs
piotmag769, line 1: I assume everything in `scarb/` is just a scarb code adaptation? If so I ain't reading that

tests/e2e/code_lens/snforge.rs
piotmag769, line 1: This is wrong

piotmag769 (changes requested):

piotmag769 (approved):

Draggu (changes requested):

tests/e2e/code_lens/snforge.rs
```diff
@@ -19,7 +19,7 @@ fn only_functions() {
     file_path = "src/lib.cairo"
 
     [[lenses]]
-    line = 2
+    line = 3
```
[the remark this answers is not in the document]
Arcticae: (will restore snapshot when i can actually run the test)

Draggu (approved):

PR #749, opened 2025-06-13
Add macro tests gor hovers and completions

### LINKED ISSUE #682: Completions in proc macro controlled code

IFoundCursedPRStackThatWillBanYou:
Close #682


piotmag769 (commented):
Add one test to each test file concerning hover and completions pls

piotmag769 (commented):
And add a test with hover on use of a new generated item i.e. function generated with macro. Same with completion (a case that should propose a new generated item)

PR #751, opened 2025-06-13
Remove relative_span field from PluginDiagnostic

maciektr:


src/lang/proc_macros/db.rs
```diff
@@ -206,8 +206,10 @@ impl SpansStabilizer {
         }
 
         for diagnostic in &mut result.diagnostics {
-            if let Some(span) = &mut diagnostic.span {
-                self.apply_original_offset_to_span(span);
+            if let Some(mut span) = diagnostic.span() {
+                self.apply_original_offset_to_span(&mut span);
```
IFoundCursedPRStackThatWillBanYou, line 210: Maybe it would be better to rework `apply_original_offset_to_span` now to return value instead but im fine with leaving it as is

IFoundCursedPRStackThatWillBanYou (approved):

src/lang/proc_macros/db.rs
maciektr: I do agree, but I personally want to avoid reworking your code :)

integraledelebesgue (approved):

PR #772, opened 2025-06-27
Fix refreshing lens + semantic tokens after macro resolution

Arcticae:


src/ide/analysis_progress.rs
```diff
@@ -79,20 +99,20 @@ impl AnalysisProgressController {
     }
 
     pub fn try_stop_analysis(&self, diagnostics_cancelled: bool) {
-        if !diagnostics_cancelled {
-            self.state.lock().unwrap().try_stop_analysis(
-                self.server_tracker.get_did_submit_procmacro_request(),
-                self.server_tracker.get_server_status(),
-            );
+        if diagnostics_cancelled {
```
Draggu, line 102: Can we move this check outside? It is akward.

src/ide/analysis_progress.rs
```diff
@@ -119,7 +152,9 @@ struct AnalysisProgressControllerState {
 
 impl AnalysisProgressControllerState {
     fn new(notifier: Notifier) -> Self {
-        Self { notifier, analysis_in_progress: false, procmacros_enabled: None }
+        let (sender, receiver) = crossbeam::channel::unbounded();
```
Draggu, line 155: Move to `AnalysisProgressUpdateChannels::new`

src/ide/analysis_progress.rs
```diff
@@ -139,19 +174,26 @@ impl AnalysisProgressControllerState {
     fn try_stop_analysis(
```
Draggu, line 174: Can we have a short comment that returned value is indicating if stopping succeeded?

src/ide/analysis_progress.rs
```diff
@@ -139,19 +174,26 @@ impl AnalysisProgressControllerState {
...
 
         self.analysis_in_progress = false;
         self.notifier.notify::<ServerStatus>(ServerStatusParams {
             event: ServerStatusEvent::AnalysisFinished,
             idle: true,
         });
+        true
     }
 
+    // WARNING: This should not be polled from outside the diagnostics thread, it may yield inaccurate results
```
Draggu, line 196: Which one? main or worker? Please make it more precise

src/lib.rs
```diff
@@ -384,6 +388,13 @@ impl Backend {
 
                     scheduler.local(Self::on_proc_macro_error);
                 }
+                recv(analysis_progress_status_receiver) -> analysis_progress_status => {
+                    let Ok(analysis_progress_status) = analysis_progress_status else { break };
+
+                    scheduler.local(|state: &mut State, _notifier: Notifier, requester: &mut Requester, _responder: Responder|
```
Draggu, line 394:
```suggestion
                    scheduler.local(|state, _notifier, requester, _responder|
```
No need to explicitly type it, inference will do the job here

src/ide/analysis_progress.rs
```diff
@@ -105,12 +125,25 @@ impl AnalysisProgressController {
     }
 }
 
+// We don't need to track starts for now
+#[derive(PartialEq)]
+pub enum AnalysisProgressStatus {
+    Stopped,
```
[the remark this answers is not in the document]
Arcticae: > are only sync tasks now 

this is not true since diagnostics can happen after this as well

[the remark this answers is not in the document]
Arcticae: I think being specific here is better for readability

src/ide/analysis_progress.rs
Arcticae: yes

PR #780, opened 2025-06-30
Propose only dot completions for `ExprBinary`s 

Arcticae:


piotmag769 (approved):

src/ide/completion/mod.rs
```diff
@@ -114,20 +114,20 @@ fn complete_ex(
...
 
     if_chain!(
-        if let Some(constructor) = node.ancestor_of_type::<ExprStructCtorCall>(db);
-        if let Some(struct_completions) = struct_constructor_completions(db, &ctx, constructor);
+        if let Some(binary_expression) = dot_expr_rhs(db, &node);
+        if let Some(dot_completions) = dot_completions(db, &ctx, binary_expression);
 
         then {
-            completions.extend(struct_completions);
+            return Some(dot_completions);
```
Draggu, line 121: This should not work like this. If we don't want some kind of completion to be showed in this context we should add negative selector in that kind. Please keep it independent

Draggu (changes requested):

Draggu (approved):
But please remove in `use`, `attr`, `params` and where it is unnecessary

PR #784, opened 2025-07-01
Highlight inline macros semantically even if same name in scope

wawel37:


integraledelebesgue (approved):

src/ide/semantic_highlighting/token_kind.rs
```diff
@@ -124,6 +124,9 @@ impl SemanticTokenKind {
         let mut expr_path_ptr = None;
 
         for node in identifier_parent.ancestors_with_self(db) {
+            if is_inline_macro(db, node) {
+                return Some(SemanticTokenKind::InlineMacro);
+            }
```
piotmag769, line 129:
```suggestion
        if identifier_parent.ancestor_of_kinds(db, [SyntaxKind::ExprInlineMacro, SyntaxKind::ItemInlineMacro]).is_some() {
            return Some(SemanticTokenKind::InlineMacro);
        }
        
        for node in identifier_parent.ancestors_with_self(db) {
```

src/ide/semantic_highlighting/token_kind.rs
```diff
@@ -271,3 +274,18 @@ fn get_resultants_and_closest_terminals(
...
+        && let Some(parent) = node.parent(db)
+        && parent.kind(db) == SyntaxKind::ExprPathInner
+        && let Some(grandparent) = parent.parent(db)
+        && grandparent.kind(db) == SyntaxKind::ExprPath
+    {
+        if let Some(great_grandparent) = grandparent.parent(db) {
+            return great_grandparent.kind(db) == SyntaxKind::ExprInlineMacro;
+        }
+    }
+    false
```
Draggu, line 290:
```suggestion
    if let Some(path_node) = node.ancestor_of_kind(SyntaxKind::ExprPath) && let Some(maybe_macro) = path_node.parent()
    {
        let kind = maybe_macro.kind(db);
        return kind == SyntaxKind::ExprInlineMacro || kind == SyntaxKind::ItemInlineMacro;
    }
    false
```

@piotmag769 version is incorrect as it will catch whole 
```
macro_call!(args)
```
instead of `macro_call` only.

Yours is to precise, if syntax definition changes even slightly here it will break, try to avoid this precision if possible to do without ambiguity

PR #786, opened 2025-07-02
Deduplicate path postfix completions

### LINKED ISSUE #710: Deduplicate equivalent completions in `path_suffix_completion`

integraledelebesgue:
Closes #710

### DIFF
--- a/src/ide/completion/path.rs
+++ b/src/ide/completion/path.rs
@@ -14,6 +14,7 @@ use lsp_types::{CompletionItem, CompletionItemKind};
 use super::helpers::completion_kind::{
     importable_completion_kind, resolved_generic_item_completion_kind,
 };
+use crate::ide::completion::CompletionItemHashable;
 use crate::lang::analysis_context::AnalysisContext;
 use crate::lang::db::AnalysisDatabase;
 use crate::lang::importer::new_import_edit;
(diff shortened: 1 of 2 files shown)


src/ide/completion/path.rs
```diff
@@ -105,6 +105,7 @@ pub fn path_suffix_completions(
                 ..CompletionItem::default()
             })
         })
+        .unique_by(|completion| serde_json::to_string(&completion).ok())
```
piotmag769, line 108: Imo better to have `CompletionItemHashable` or sth

integraledelebesgue: It requires a lot of excessive work to include all non-hashable data from `CompletionItem` (including the private fields). I think it's better to use JSON here

piotmag769 (approved):

src/ide/completion/path.rs
Draggu: You know that `CompletionItemHashable` already exists and does this json trick right?

integraledelebesgue: Didn't notice, thank you!

piotmag769 (approved):

Draggu (approved):

Draggu (approved):

PR #800, opened 2025-07-07
Introduce task without mutable access

Draggu:
This allows us to get access to requester/responder without mut access to state.
This in turn allows us to skip hooks. It is necessary for next PR to not fall into infinite loop.

---

**Stack**:
- #801
- #800 ⬅

### DIFF
--- a/src/server/schedule/mod.rs
+++ b/src/server/schedule/mod.rs
@@ -17,7 +17,8 @@ mod task;
 pub mod thread;
 
 pub(super) use self::task::BackgroundSchedule;
-pub use self::task::{SyncTask, Task};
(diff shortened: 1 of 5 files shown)


src/server/schedule/mod.rs
```diff
@@ -89,12 +95,22 @@ impl<'s> Scheduler<'s> {
         }
     }
 
+    /// Dispatches a local mutable `task`.
```
piotmag769, line 98:
```suggestion
    /// Dispatches a local task with access to the mutable state.
```

piotmag769 (approved):

PR #805, opened 2025-07-09
Change apply responses trigger

Draggu:
In fact `find_primary_files` does all expansion, use it as a trigger for potential responses applyings. This avoids busy waiting on main thread, also allows us to make expansion block shorter

---

**Stack**:
- #805 ⬅
- #804


src/lang/proc_macros/controller.rs
```diff
@@ -431,28 +438,23 @@ pub struct ProcMacroChannels {
...
-    pub response_receiver: Receiver<()>,
-
-    // A single element queue is used to notify when the response queue is pushed.
-    pub response_sender: Sender<()>,
-
     // A single element queue is used to notify when client occurred an error.
     pub error_receiver: Receiver<()>,
+
+    // A single element queue is used to notify when responses should be applied.
+    pub poll_response_receiver: Receiver<()>,
```
piotmag769, line 445:
```suggestion
    pub poll_responses_receiver: Receiver<()>,
```

src/lang/proc_macros/response_poll.rs
```diff
@@ -0,0 +1,32 @@
...
+
+        thread::Builder::new(ThreadPriority::Worker)
+            .name("cairo-ls:pms-response-poll".into())
+            .spawn(move || this.event_loop())
+            .expect("failed to spawn pms response poll thread")
+    }
+
+    fn event_loop(self) {
+        loop {
+            let _ = match self.found_files_receiver.recv_timeout(Duration::from_secs(3)) {
```
piotmag769, line 25: Explain that this is the "rate limiter" in comment and how it works (why we are waiting on `found_files_receiver`)

src/lib.rs
```diff
@@ -386,14 +380,17 @@ impl Backend {
...
-                        scheduler.local_mut(Self::on_proc_macro_response);
-                    } else {
-                        let _ = proc_macro_channels.response_sender.try_send(());
-                    }
+                    scheduler.local_with_condition(|state| {
+                        if let ServerStatus::Starting(client) | ServerStatus::Ready(client) = state.db.proc_macro_server_status() {
+                            client.available_responses().len() != 0
+                        } else {
+                            false
+                        }
```
piotmag769, line 392: Can we extract to function to name it?

src/lang/diagnostics/mod.rs
```diff
@@ -145,6 +157,8 @@ impl DiagnosticsControllerThread {
         let (state, primary_snapshots, secondary_snapshots) = state_snapshots.split();
 
         let primary_set = find_primary_files(&state.db, &state.open_files);
+        // Event meaning that `find_primary_files` finished it's work.
+        let _ = self.found_files_sender.send(());
```
piotmag769, line 161: Send after `find_secondary_files` to account for the case where `open_files` are empty. If it makes the perf worse then we can just ignore it, this case is rly rare

maciektr (approved):

src/lib.rs
```diff
@@ -426,6 +416,16 @@ impl Backend {
         state.proc_macro_controller.handle_error(&mut state.db, &state.config);
     }
 
+    fn proc_macro_response_check(state: &State) -> bool {
```
piotmag769, line 419: `any_proc_macro_response_available`

PR #806, opened 2025-07-09
Fix proc macro cache path

Draggu:


piotmag769 (commented):
Add error logging when you cannot write to cache too

piotmag769 (approved):

maciektr (approved):

PR #834, opened 2025-07-22
Add `scarb execute` code lens

Arcticae:
Example of how it looks:
image

### DIFF
--- a/src/ide/code_lens/executables.rs
+++ b/src/ide/code_lens/executables.rs
@@ -1,6 +1,6 @@
 use crate::ide::code_lens::{
-    AnnotatedNode, FileCodeLens, LSCodeLens, LSCodeLensInterface, collect_functions_with_attrs,
-    get_original_node_and_file, make_lens_args,
+    AnnotatedNode, CodeLensBuilder, CodeLensInterface, CodeLensProvider, LSCodeLens,
+    collect_functions_with_attrs, get_original_node_and_file, make_lens_args,
 };
 use crate::lang::db::AnalysisDatabase;
 use crate::lang::lsp::LsProtoGroup;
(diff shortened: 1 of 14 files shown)


wawel37 (approved):

src/ide/code_lens/executables.rs
```diff
@@ -0,0 +1,178 @@
...
+
+#[derive(PartialEq)]
+struct ExecutableLensPayload {
+    command: String,
+}
+
+#[derive(PartialEq)]
+pub struct ExecutableCodeLens {
+    lens: CodeLens,
+    payload: ExecutableLensPayload,
```
piotmag769, line 31:
```suggestion
    command: String,
```
You don't use this field anywhere else anyways

src/ide/code_lens/mod.rs
```diff
@@ -256,49 +298,113 @@ pub struct FileChange {
...
+    let [Value::String(function_full_path), Value::String(url)] = args else {
+        return None;
+    };
     let url: Url = url.parse().ok()?;
 
-    Some((index, url))
+    Some((function_full_path.clone(), url))
+}
+
+pub struct AnnotatedNode {
```
piotmag769, line 366: Doesn't need to be pub if used only in child modules

src/ide/code_lens/executables.rs
```diff
@@ -0,0 +1,178 @@
...
+                return None;
+            };
+            Some(cu.target.name.clone())
+        } else {
+            None
+        }
+    })
+}
+
+fn precise_cu_available(
```
piotmag769, line 131: This is not what this function does. It looks for target instead. Btw we shouldn't even pass CUs here, just targets

src/ide/code_lens/executables.rs
```diff
@@ -0,0 +1,178 @@
...
+    function_full_path: &String,
+    compilation_units: Vec<CompilationUnitMetadata>,
+) -> bool {
+    compilation_units.iter().any(|cu| {
+        if cu.target.kind == "executable" {
+            let Some(params_obj) = cu.target.params.as_object() else { return false };
+            let Some(func) = params_obj.get("function") else { return false };
+            let Some(func_name) = func.as_str() else { return false };
```
piotmag769, line 139: chain if let, it is available in new Rust

src/ide/code_lens/mod.rs
```diff
@@ -256,49 +298,113 @@ pub struct FileChange {
...
+    pub full_path: String,
+    pub attribute_ptr: SyntaxStablePtrId,
+}
+/// Collects functions with given attributes on them
+/// Returns tuples of (full path, pointer to found attribute)
+pub fn collect_functions_with_attrs(
+    db: &AnalysisDatabase,
+    module: ModuleId,
+    attributes: Vec<&str>,
```
piotmag769, line 375:
```suggestion
    attributes: [&str],
```

src/ide/code_lens/executables.rs
```diff
@@ -0,0 +1,178 @@
...
+    CodeLens { range, command: Some(command), data: None }
+}
+
+fn collect_executable_functions(db: &AnalysisDatabase, module: ModuleId) -> Vec<AnnotatedNode> {
+    collect_functions_with_attrs(db, module, Vec::from(PLAIN_EXECUTABLES))
+}
+
+fn get_executable_lens_position(
+    db: &AnalysisDatabase,
+    function_ptr: SyntaxStablePtrId,
```
piotmag769, line 170: It is an attr ptr. You don't need half of this function's logic

piotmag769 (dismissed):
Add test with both `#[executable]` and `#[test]`

src/ide/code_lens/mod.rs
```diff
@@ -3,25 +3,65 @@ use std::collections::hash_map::Entry;
...
+use itertools::Itertools;
+use lsp_types::request::CodeLensRefresh;
+use lsp_types::{CodeLens, Url};
+use serde::{Deserialize, Serialize};
+use serde_json::{Number, Value};
 
+mod executables;
 mod tests;
 
+trait LSCodeLensInterface {
```
piotmag769, line 32:
```suggestion
trait CodeLensInterface {
```

tests/e2e/code_lens/executable.rs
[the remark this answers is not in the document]
Arcticae: Scarb does not support inline modules here, how would this work anyway? 🤔

PR #830, opened 2025-07-21
Swap the database after applying a certain number of mutations

integraledelebesgue:
## Changes
* Swapper triggers every `15 minutes` instead of `5`
* Swap also occurs after applying `5000 mutations` to the database
* Numbers are educated guesses and can be tuned in future PRs

## Stack
* ...
* https://github.com/software-mansion/cairols/pull/835
* https://github.com/software-mansion/cairols/pull/830 ◀️

PR #835, opened 2025-07-22
Track only the active analysis time in `AnalysisDatabaseSwapper`

### LINKED ISSUE #823: Do not count idle time to swapper
You turn on the LS/IDE, go make a cup of tea and then swap occurs, even though you did nothing

We should somehow exclude idle time - swap only when active for a certain time. 
Proposition for implementation: stop the swapper "stopwatch" when analysis was not in progress for some time. Swap when the stopwatch reaches a certain time value.

integraledelebesgue:
Closes #823

## Changes
* `AnalysisDatabaseSwapper` belongs to `MetaState` instead of `State`
* `AnalysisFinished` becomes `AnalysisStatus` enum which indicates the start and the end of the analysis
* `AnalysisDatabaseSwapper` uses `Stopwatch` mechanic to track the analysis time while skipping the idle

## Stack
* ...
* https://github.com/software-mansion/cairols/pull/835 ◀️ 
* https://github.com/software-mansion/cairols/pull/830

### DIFF
--- a/src/lang/db/swapper.rs
+++ b/src/lang/db/swapper.rs
@@ -1,5 +1,6 @@
 use std::collections::HashSet;
 use std::fmt::Display;
+use std::mem;
 use std::panic::{AssertUnwindSafe, catch_unwind};
 use std::sync::Arc;
 use std::time::{Duration, SystemTime};
@@ -58,7 +59,7 @@ impl Display for SwapReason {
 /// It is expected that diagnostics will be refreshed on it as quickly as possible, otherwise
 /// the entire workspace would be recompiled at an undetermined time leading to bad UX delays.
 pub struct AnalysisDatabaseSwapper {
-    last_replace_time: SystemTime,
+    stopwatch: Stopwatch,
     mutations_since_last_replace: u64,
     db_replace_min_interval: Duration,
     db_replace_min_mutations: u64,
@@ -67,7 +68,7 @@ pub struct AnalysisDatabaseSwapper {
 impl Default for AnalysisDatabaseSwapper {
     fn default() -> Self {
         Self {
-            last_replace_time: SystemTime::now(),
+            stopwatch: Stopwatch::default(),
             mutations_since_last_replace: 0,
             db_replace_min_interval: env_config::db_replace_interval(),
             db_replace_min_mutations: env_config::db_replace_mutations(),
@@ -78,6 +79,19 @@ impl Default for AnalysisDatabaseSwapper {
 impl AnalysisDatabaseSwapper {
     pub fn register_mutation(&mut self) {
         self.mutations_since_last_replace += 1;
+    }
+
+    pub fn start_stopwatch(&mut self) {
+        self.stopwatch.start();
+        trace!("Stopwatch started!");
+    }
+
+    pub fn stop_stopwatch(&mut self) {
+        self.stopwatch.stop();
+        trace!(
+            "Stopwatch stopped! Total elapsed time: {}s",
+            self.stopwatch.total_elapsed_time.as_secs()
+        );
     }
 
     /// Checks for the swap criteria and swaps the database if they have been met.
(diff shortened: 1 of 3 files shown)


src/server/schedule/mod.rs
```diff
@@ -154,6 +154,15 @@ impl<'s> Scheduler<'s> {
         self.dispatch(Task::local_with_precondition(precondition_func, mut_func));
     }
 
+    /// Immedietely executes a function which mutates the underlying [`State`], bypassing the dispatch phase.
+    ///
+    /// # Warning
+    /// Use only for lightweight, instant actions and when `local_mut` task is not suitable.
+    /// Mind the potential race conditions caused by non-dispatched mutations.
+    pub fn instant(&mut self, func: impl FnOnce(&mut State) + 's) {
```
Draggu, line 162: Remove ,this is a footgun. It would be better to have second state that is independent from this.

src/lib.rs
```diff
@@ -394,11 +394,22 @@ impl Backend {
                     scheduler.local_mut(Self::on_proc_macro_error);
                 }
                 recv(analysis_progress_status_receiver) -> analysis_progress_status => {
-                    let Ok(AnalysisFinished) = analysis_progress_status else { break };
-
-                    scheduler.local(|state, _notifier, requester, _responder|
-                        Self::on_stopped_analysis(state, requester)
-                    );
+                    let Ok(analysis_status) = analysis_progress_status else { break };
```
Draggu, line 397: Not a big fan of doing more work on main loop than necessary, AnalysisThread can track analysis time in some atomic variable, it would be enough

src/lang/db/swapper.rs
```diff
@@ -57,17 +57,42 @@ impl Display for SwapReason {
...
+    fn start(&mut self) {
+        self.start_time = Some(SystemTime::now());
+    }
+
+    fn stop(&mut self) {
+        let Some(start_time) = self.start_time else {
+            return;
+        };
+
+        let elapsed_time = SystemTime::now().duration_since(start_time).unwrap_or_default();
```
Draggu, line 82: Please log if it failed, it can only happen if time went backward. It is better to know that fact

Draggu (changes requested):

src/lang/db/swapper.rs
```diff
@@ -57,17 +57,42 @@ impl Display for SwapReason {
...
+        self.total_elapsed_time += elapsed_time;
```
Arcticae, line 83: We should set `start_time` here to `None` in order to avoid double `stop` (and incorrect summing up of times). An assert would be also nice in order to make sure we don't didn't actually code anything that remotely tries to do this.

src/lang/db/swapper.rs
```diff
@@ -57,17 +57,42 @@ impl Display for SwapReason {
...
+    }
```
Arcticae, line 84: We should set `start_time` to `None` and `assert!` at the beginning to avoid incorrect summing up numbers + actually avoiding developer errors

src/lib.rs
Arcticae: Do we need this kind of resolution in time tracking? How much time would be "lost" if the actual tracking would happen via a task?

Arcticae (changes requested):

Arcticae (approved):
Gud but pls lock the whole metaState - rebase with https://github.com/software-mansion/cairols/pull/842

src/lib.rs
```diff
@@ -339,11 +339,28 @@ impl Backend {
...
                 recv(analysis_progress_status_receiver) -> analysis_progress_status => {
-                    let Ok(AnalysisFinished) = analysis_progress_status else { break };
-
-                    scheduler.local(|state, _, _notifier, requester, _responder|
-                        Self::on_stopped_analysis(state, requester)
-                    );
+                    let Ok(analysis_status) = analysis_progress_status else { break };
+
+                    match analysis_status {
+                        AnalysisStatus::Started => {
```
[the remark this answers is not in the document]
integraledelebesgue: If it was so, the analysis thread would need to modify the meta-state, which would be used in the main loop anyway. The current logic seems more tight and clear to me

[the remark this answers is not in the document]
integraledelebesgue: Also, the stopwatch should be attached to the swapper because it needs to be zeroed between swaps. Counting the time on the analysis thread would complicate this

Draggu (approved):

Issue #799, opened 2025-07-07
Abort some operations while diagnostics are not resolved


piotmag769:
Some operations like semantic highlighting (and others that have to use semantics which are resolved during calculating diagnostics) may lock salsa for extended periods of time due to them resolving very complex queries. We would prefer these queries to be (at least partially) calculated on our multiple diagnostic threads. Therefore we should abort some requests (meaning return `None` or last returned value immediately) if diagnostics are still being calculated.

PR #842, opened 2025-07-25
Implement meta LS state 

Arcticae:
It's needed for #823 and #799 in order to keep information about swapper state and analysis progress without launching the hooks

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 use std::default::Default;
 use std::ops::{Deref, DerefMut};
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use lsp_types::{ClientCapabilities, Url};
 use salsa::ParallelDatabase;
@@ -86,10 +86,12 @@ impl State {
     }
 }
 
+#[derive(Default)]
+pub struct MetaStateInner {}
+
 /// State keeps information about LS state (swapper, analysis state or other internal info)
-/// Mutations don't influence refreshing of diagnostics and hooks etc.
-#[derive(Default, Clone)]
-pub struct MetaState {}
+/// Mutations of this struct are allowed in background tasks and do not trigger hooks.
+pub type MetaState = Arc<Mutex<MetaStateInner>>;
 
 /// Readonly snapshot of Language server state.
 pub struct StateSnapshot {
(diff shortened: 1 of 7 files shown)


src/state.rs
```diff
@@ -86,6 +86,12 @@ impl State {
     }
 }
 
+// TODO: Implement
+/// State keeps information about LS state (swapper, analysis state or other internal info)
+/// Mutations don't influence refreshing of diagnostics and hooks etc.
+#[derive(Default, Clone)]
+pub struct MetaState {}
```
Draggu, line 93: Make it `Arc` wrapped by default everywhere, this way we don't have to keep each field with its own

src/state.rs
```diff
@@ -86,6 +86,12 @@ impl State {
     }
 }
 
+// TODO: Implement
+/// State keeps information about LS state (swapper, analysis state or other internal info)
```
piotmag769, line 90:
```suggestion
/// Information about LS state itself (swapper, analysis state or other internal info)
```

src/state.rs
```diff
@@ -86,6 +86,12 @@ impl State {
...
+/// Mutations don't influence refreshing of diagnostics and hooks etc.
```
piotmag769, line 91:
```suggestion
/// Mutations of this struct do not trigger hooks.
```

src/state.rs
```diff
@@ -86,6 +86,12 @@ impl State {
...
+#[derive(Default, Clone)]
```
integraledelebesgue, line 92: `Clone` is not necessary since we will wrap the only instance with `Arc`

integraledelebesgue (approved):
Looks good, should be fine for #823

integraledelebesgue (approved):

src/state.rs
```diff
@@ -86,6 +86,15 @@ impl State {
     }
 }
 
+// TODO: Implement
+
+#[derive(Default)]
+pub struct MetaStateInner {}
+
+/// Information about LS state itself (swapper, analysis state or other internal info).
+/// Mutations of this struct do not trigger hooks.
```
piotmag769, line 95:
```suggestion
/// Mutations of this struct are allowed in background tasks and do not trigger hooks.
```

src/state.rs
```diff
@@ -86,6 +86,15 @@ impl State {
     }
 }
 
+// TODO: Implement
```
piotmag769, line 89: Remove

piotmag769 (approved):

PR #845, opened 2025-07-28
Remove old linter plugin and use query group from linter

wawel37:


src/ide/introspection/crates.rs
```diff
@@ -39,7 +37,6 @@ struct CrateView {
     name: SmolStr,
     source_paths: Vec<PathBuf>,
     settings: CrateSettings,
-    linter_configuration: LinterConfiguration,
```
Draggu, line 42: Why we can't see linter config here now?

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -60,6 +84,24 @@ impl FilesDiagnostics {
                     db.module_lowering_diagnostics(module_id).unwrap_or_default().get_all()
                 }),
             );
+            if config.enable_linter {
```
Draggu, line 87: We should skip crates with disabled linter here, also skip deps. I see we have this in `linter_params`. Is it done internally in `db.linter_diagnostics`?

src/lang/diagnostics/refresh.rs
```diff
@@ -7,17 +7,22 @@ use cairo_lang_utils::LookupIntern;
...
 
 /// Refresh diagnostics and send diffs to the client.
 #[tracing::instrument(skip_all)]
+#[allow(clippy::too_many_arguments)]
 pub fn refresh_diagnostics(
     db: &AnalysisDatabase,
+    config: &Config,
+    config_registry: &ConfigsRegistry,
     batch: Vec<FileId>,
     trace_macro_diagnostics: bool,
```
Draggu, line 27: It is in `Config`, now we can stop passing it and use `config.trace_macro_diagnostics` instead. Can be separate PR

src/lang/diagnostics/refresh.rs
```diff
@@ -42,15 +49,20 @@ pub fn refresh_diagnostics(
 /// I.e, if diagnostics are updated on the server side they MUST be sent successfully to the
 /// client (and vice-versa).
 #[tracing::instrument(skip_all, fields(url = tracing_file_url(db, root_on_disk_file)))]
+#[allow(clippy::too_many_arguments)]
 fn refresh_file_diagnostics(
     db: &AnalysisDatabase,
+    config: &Config,
+    config_registry: &ConfigsRegistry,
     root_on_disk_file: FileId,
     trace_macro_diagnostics: bool,
```
Draggu, line 58: vide supra

src/project/model/mod.rs
```diff
@@ -183,10 +172,4 @@ impl ProjectModel {
 
         self.loaded_workspaces.insert(workspace_dir.to_path_buf(), workspace_crates);
     }
-
-    fn is_from_scarb_cache(&self, crate_root_path: &Path) -> bool {
```
Draggu, line 187: Like described above, we should NOT run linter for cache located crates

Draggu (changes requested):

src/lang/diagnostics/file_diagnostics.rs
wawel37: I don't know if LS can decide upon if something is a dependency. If it doesn't, neither can linter as it's get only the moduleId

src/lang/diagnostics/refresh.rs
wawel37: I don't get it how `trace_macro_diagnostics` is related to this use case.

wawel37: Lets do it in separate PR then

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -34,20 +41,41 @@ pub struct FilesDiagnostics {
...
+        let linter_params = LinterDiagnosticParams {
+            only_generated_files: false,
+            tool_metadata: config_registry
+                .config_for_file(root_path)
+                .map_or_else(CairoLintToolMetadata::default, |config| config.lint.clone()),
+        };
 
         let (files_to_process, modules_to_process) =
-            db.file_and_subfiles_with_corresponding_modules(root_on_disk_file)?;
+            <AnalysisDatabase as LsSemanticGroup>::file_and_subfiles_with_corresponding_modules(
```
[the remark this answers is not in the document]
wawel37: both `LsSemanticGroup` and `LinterGroup` got this method. It will removable after upstreaming parts of ls logic from ls and linter to separate crate

tests/e2e/scarb/scarb_toml_change/snapshots/e2e__scarb__scarb_toml_change__removing_dependency__removing_dependency.snap
```diff
@@ -45,7 +45,6 @@ analyzed_crates = '''
       "user_defined_inline_macros": false
     }
   },
-  "linter_configuration": "Off",
```
piotmag769, line 48: Why? We can easily retrieve this information: is dep from scarb cache + is linter on + configs registry

src/ide/code_actions/cairo_lint.rs
```diff
@@ -55,3 +84,18 @@ pub fn cairo_lint(db: &AnalysisDatabase, ctx: &AnalysisContext<'_>) -> Option<Ve
...
+    config_registry: &ConfigsRegistry,
+) -> CairoLintToolMetadata {
+    if let Ok(module_file_id) = ctx.module_file_id.file_id(db)
+        && let FileLongId::OnDisk(file_id) = module_file_id.lookup_intern(db)
+        && let Some(file_config) = config_registry.config_for_file(&file_id)
+    {
+        return file_config.lint.clone();
+    }
+
+    Default::default()
```
piotmag769, line 100:
```suggestion
    if let Ok(module_file_id) = ctx.module_file_id.file_id(db)
        && let FileLongId::OnDisk(file_id) = module_file_id.lookup_intern(db)
        && let Some(file_config) = config_registry.config_for_file(&file_id)
    {
        file_config.lint.clone()
    } else {
        Default::default()
    }
```

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -34,20 +41,41 @@ pub struct FilesDiagnostics {
...
+    ) -> Option<Self> {
         let root_on_disk_file_url = db.url_for_file(root_on_disk_file)?;
 
         let mut semantic_file_diagnostics: Vec<SemanticDiagnostic> = vec![];
         let mut lowering_file_diagnostics: Vec<LoweringDiagnostic> = vec![];
         let mut parser_file_diagnostics: Vec<ParserDiagnostic> = vec![];
+        let mut linter_file_diagnostics: Vec<SemanticDiagnostic> = vec![];
+
+        let root_path_string = root_on_disk_file.full_path(db);
+        let root_path = Path::new(root_path_string.as_str());
```
piotmag769, line 65: Merge into one line

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -34,20 +41,41 @@ pub struct FilesDiagnostics {
...
+        let corelib_context = CorelibContext::new(db);
```
piotmag769, line 66: What is this? From the name I cannot tell what it is actually for

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -34,20 +41,41 @@ pub struct FilesDiagnostics {
...
+        let linter_params = LinterDiagnosticParams {
+            only_generated_files: false,
```
piotmag769, line 68: What does this flag do? Seems weird to even be able to set it to true

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -60,6 +88,24 @@ impl FilesDiagnostics {
                     db.module_lowering_diagnostics(module_id).unwrap_or_default().get_all()
                 }),
             );
+            if config.enable_linter && !scarb_toolchain.is_from_scarb_cache(root_path) {
```
piotmag769, line 91: Add a comment why (about scarb cache). I know it wasn't there but it is useful.

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -34,20 +41,41 @@ pub struct FilesDiagnostics {
     pub parser: Diagnostics<ParserDiagnostic>,
     pub semantic: Diagnostics<SemanticDiagnostic>,
     pub lowering: Diagnostics<LoweringDiagnostic>,
+    pub linter: Diagnostics<SemanticDiagnostic>,
```
piotmag769, line 44: No need to have a separate field for that - push it to `semantic` instead

src/ide/code_actions/cairo_lint.rs
```diff
@@ -1,20 +1,49 @@
...
+) -> Option<Vec<CodeAction>> {
+    let linter_params = LinterDiagnosticParams {
+        only_generated_files: false,
+        tool_metadata: get_linter_tool_metadata(db, ctx, config_registry),
+    };
 
-pub fn cairo_lint(db: &AnalysisDatabase, ctx: &AnalysisContext<'_>) -> Option<Vec<CodeAction>> {
-    let diags = db.module_semantic_diagnostics(ctx.module_file_id.0).ok()?;
+    let module_id = ctx.module_file_id.0;
+    let semantic_diags = db.module_semantic_diagnostics(module_id).ok()?;
```
piotmag769, line 30: Why do you need to collect it if you won't ever generate fixes for non-lint diagnostics?

src/lang/diagnostics/file_diagnostics.rs
wawel37: `root_on_disk_file.full_path(db)` is temporary and dropped while borrowed. Cannot move it to single line

src/ide/code_actions/cairo_lint.rs
wawel37: We actually need this, because linter fixes the unused import diags, which come from the semantic diags. Will add a comment why we do that

src/lang/diagnostics/file_diagnostics.rs
```diff
@@ -34,20 +41,41 @@ pub struct FilesDiagnostics {
...
 }
 
 impl FilesDiagnostics {
     /// Collects all diagnostics kinds by processing an on disk `root_on_disk_file` together with
     /// virtual files that are its descendants.
-    pub fn collect(db: &AnalysisDatabase, root_on_disk_file: FileId) -> Option<Self> {
+    pub fn collect(
+        db: &AnalysisDatabase,
+        config: &Config,
```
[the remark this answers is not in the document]
wawel37: We passing the config by a ref, so i don't think it's a bad idea here. If we cloned whole object, yup, that would be bad. But just a ref is not a problem imo

src/ide/introspection/crates.rs
```diff
@@ -86,26 +99,21 @@ enum LinterConfiguration {
...
-        config.insert(
-            "include_compiler_generated_files".to_owned(),
-            plugin.include_compiler_generated_files(),
-        );
+        }
 
-        Self::On(config)
+        Self::On(
+            configs_registry.config_for_file(root_path).map_or_else(Default::default, |config| {
+                config.lint.clone().into_iter().collect::<HashMap<_, _>>()
```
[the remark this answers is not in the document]
wawel37: actually not, beucase orderedhashmap cannot be collected directly

PR #850, opened 2025-07-29
Catch panics coming from LSP request handlers

### LINKED ISSUE #814: Don't show annoying error pop-ups
There are a few options here (not all listed, to be thought about):
1. Re-evaluate which notifications we show to the user, and which are shown in status bar (keep current approach)
2. Debounce the errors and show a notification
3. Hide all notifications by default and only show them in status-bar/console.

integraledelebesgue:
Closes #814

### DIFF
--- a/src/server/routing/handlers.rs
+++ b/src/server/routing/handlers.rs
@@ -5,6 +5,8 @@
 // | Commit: 46a457318d8d259376a2b458b3f814b9b795fe69    |
 // +-----------------------------------------------------+
 
+use std::panic::{AssertUnwindSafe, catch_unwind, resume_unwind};
+
 use cairo_lang_filesystem::db::{FilesGroup, FilesGroupEx, PrivRawFileContentQuery};
 use lsp_types::notification::{
     DidChangeConfiguration, DidChangeTextDocument, DidChangeWatchedFiles, DidCloseTextDocument,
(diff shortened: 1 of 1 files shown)


Draggu (approved):

piotmag769 (approved):
Bjutiful animal-like code

src/server/routing/handlers.rs
```diff
@@ -86,7 +88,11 @@ impl BackgroundDocumentRequestHandler for CodeActionRequest {
         _notifier: Notifier,
         params: CodeActionParams,
     ) -> Result<Option<CodeActionResponse>, LSPError> {
-        Ok(ide::code_actions::code_actions(params, &snapshot.db))
+        Ok(catch_unwind(AssertUnwindSafe(|| ide::code_actions::code_actions(params, &snapshot.db)))
+            .unwrap_or_else(|_| {
+                error!("CodeActionRequest handler panicked");
```
piotmag769, line 93: Please log the errors though

piotmag769 (changes requested):
You catch cancellations by accident this way xD please resume unwind if it is cancelled

wawel37 (approved):

integraledelebesgue:
> You catch cancellations by accident this way xD please resume unwind if it is cancelled. Reuse existing functions from our codebase

Done, please TAL

src/server/routing/handlers.rs
```diff
@@ -481,11 +552,20 @@ impl BackgroundDocumentRequestHandler for CodeLensRequest {
...
-        ))
+        Ok(catch_unwind(AssertUnwindSafe(|| {
+            snapshot.code_lens_controller.code_lens(
+                params.text_document.uri,
+                &snapshot.db,
+                &snapshot.config,
+            )
+        }))
+        .unwrap_or_else(|err| {
+            if err.downcast_ref::<salsa::Cancelled>().is_some() {
```
piotmag769, line 563: You won't catch the thing where salsa is bugged and it doesn't work. That's why I said to use our function - `is_cancelled`

PR #854, opened 2025-07-30
Inject `MetaState` to `Scheduler`

integraledelebesgue:


Arcticae (changes requested):
I don't see a reason to do this TBH - why do you need this?

integraledelebesgue:
> I don't see a reason to do this TBH - why do you need this?

I need this in https://github.com/software-mansion/cairols/pull/855 to pass the event sender into the swapper in a clean way

PR #855, opened 2025-07-30
Notify `AnalysisProgressThread` about database swaps

integraledelebesgue:


src/ide/analysis_progress.rs
```diff
@@ -220,6 +225,12 @@ impl AnalysisProgressThread {
 
                     pms_status = new_pms_status;
                 }
+                AnalysisEvent::DatabaseSwap => {
```
Draggu, line 228: You actually need whole logic from `Mutation` i think it would be best to trick them as equal in this match
```suggestion
                AnalysisEvent::Mutation | AnalysisEvent::DatabaseSwap => {
```

Draggu (changes requested):

Draggu (approved):

Issue #396, opened 2025-03-01
Support all possible elements in autoimports.


Draggu:
Currently we ignore "nested" elements like trait/impl fn/const/type

THenry14:
@Draggu please improve the description

PR #857, opened 2025-07-30
Add trait items to path prefix completions

piotmag769:


integraledelebesgue (approved):

wawel37 (approved):

Draggu (approved):

src/ide/completion/path.rs
```diff
@@ -181,3 +164,37 @@ pub fn path_prefix_completions(
         _ => vec![],
     })
 }
+
+fn trait_items_completions(db: &AnalysisDatabase, trait_id: TraitId) -> Vec<CompletionItem> {
```
Arcticae, line 168: Does it make sense to propose all of those? 
i.e.: why would i want to have impl names or types proposed here?

piotmag769:
>What is this and why do we need this

Forgot to add tests, will do. It is related to #396. We need this so completion of trait items would work. But now I realised I wrote it a wrong way xD

PR #864, opened 2025-08-01
Send execute in terminal conditionally

### LINKED ISSUE #862: Code Lens `cairo/executeInTerminal` custom nitification capabilities
Implement custom client capabilities for `cairo/executeInTerminal`. This way we can detect if user client is supporting `cairo/executeInTerminal` and make alternative move like not showing lenses or executing command on LS side. This is important in editors others than VSC as these currently does not support or custom notification leading to code lens execution being no-op

piotmag769:
Closes #862


src/ide/code_lens/tests.rs
```diff
@@ -94,10 +96,18 @@ impl CodeLensProvider for TestCodeLensProvider {
...
-            cwd: state.project_controller.configs_registry().manifest_dir_for_file(&file_path)?,
-            command,
-        });
+        let cwd = state.project_controller.configs_registry().manifest_dir_for_file(&file_path)?;
+        if state.client_capabilities.execute_in_terminal_support() {
+            notifier.notify::<ExecuteInTerminal>(ExecuteInTerminalParams { cwd, command });
+        } else {
+            notifier.notify::<ShowMessage>(ShowMessageParams {
+                typ: MessageType::INFO,
+                message: format!(
```
Arcticae, line 105: We should watch out in PRs in the future so we don't duplicate this logic (should be common for all operations that are not supported)

Arcticae (approved):

src/ide/code_lens/tests.rs
piotmag769: Idk how to deduplicate it tbh, but we'll see

PR #871, opened 2025-08-04
Add label detail if multiple completions with same label

wawel37:
image

### DIFF
--- a/src/ide/completion/path.rs
+++ b/src/ide/completion/path.rs
@@ -1,5 +1,3 @@
-use std::collections::HashSet;
-
 use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_defs::ids::NamedLanguageElementId;
 use cairo_lang_semantic::db::SemanticGroup;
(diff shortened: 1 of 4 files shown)


src/ide/completion/path.rs
```diff
@@ -98,12 +100,33 @@ pub fn path_suffix_completions(
...
+        .map(|item| &item.label)
+        .filter(|label| completions_vec.iter().filter(|item| &item.label == *label).count() > 1)
+        .cloned()
+        .collect();
+
+    for completion in &mut completions {
+        if !duplicate_labels.contains(&completion.label) {
+            completion.label_details = None;
+        }
+    }
```
integraledelebesgue, line 127: Please consider using `Itertools::counts` instead of counting manually:
```suggestion
    let label_counts = completions.iter().map(|item| item.label.clone()).counts();
    for completion in &mut completions {
        if label_counts[&completion.label] == 1 {
            completion.label_details = None;
        }
    }
```

integraledelebesgue (approved):

Arcticae (approved):

src/ide/completion/path.rs
```diff
@@ -97,12 +97,26 @@ pub fn path_suffix_completions<'db>(
...
-        .collect()
+        .collect();
+
+    // Remove path label_details from all completions, that are NOT duplicated.
+    let label_counts = completions.iter().map(|item| item.label.clone()).counts();
+    for completion in &mut completions {
+        if label_counts[&completion.label] == 1 {
+            completion.label_details = None;
+        }
+    }
```
piotmag769, line 117: Do we wanna do that tbh? I think the paths are beneficial actually, you can see where the item is from

PR #876, opened 2025-08-06
Add a query for generic parameters of an arbitrary `ModuleItem`

integraledelebesgue:
## Stack
* https://github.com/software-mansion/cairols/pull/895
* https://github.com/software-mansion/cairols/pull/878
* https://github.com/software-mansion/cairols/pull/877
* https://github.com/software-mansion/cairols/pull/881
* https://github.com/software-mansion/cairols/pull/876 ◀️

### DIFF
--- a/src/lang/db/semantic.rs
+++ b/src/lang/db/semantic.rs
@@ -10,11 +10,11 @@ use cairo_lang_defs::ids::{
 };
 use cairo_lang_filesystem::db::{get_parent_and_mapping, translate_location};
 use cairo_lang_filesystem::ids::{FileId, FileLongId};
-use cairo_lang_semantic::Binding;
 use cairo_lang_semantic::db::SemanticGroup;
 use cairo_lang_semantic::expr::pattern::QueryPatternVariablesFromDb;
 use cairo_lang_semantic::items::function_with_body::SemanticExprLookup;
 use cairo_lang_semantic::lookup_item::LookupItemEx;
+use cairo_lang_semantic::{Binding, GenericParam};
 use cairo_lang_syntax::node::helpers::GetIdentifier;
 use cairo_lang_syntax::node::kind::SyntaxKind;
 use cairo_lang_syntax::node::{SyntaxNode, TypedSyntaxNode, ast};
@@ -23,6 +23,7 @@ use cairo_lang_utils::{Intern, Upcast};
 
 use super::LsSyntaxGroup;
 
+#[allow(dead_code)] // Used later in the stack
 #[cairo_lang_proc_macros::query_group(LsSemanticDatabase)]
 pub trait LsSemanticGroup:
     SemanticGroup + for<'db> Upcast<'db, dyn SemanticGroup> + LsSyntaxGroup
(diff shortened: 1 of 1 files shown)


src/lang/db/semantic.rs
```diff
@@ -152,6 +153,13 @@ pub trait LsSemanticGroup:
     ///
     /// Therefore for `FooTrait` from file 1, `FooTrait` from file 1 and `FooTrait` from file 2 are returned.
     fn get_node_resultants<'db>(&'db self, node: SyntaxNode<'db>) -> Option<Vec<SyntaxNode<'db>>>;
+
+    /// Returns generic parameters defined by the item.
+    /// Items that do not introduce any generics (use statements, modules etc.) yield an empty vector.
+    fn item_generic_params<'db>(
+        &'db self,
+        lookup_item: LookupItemId<'db>,
+    ) -> Vec<GenericParam<'db>>;
```
wawel37, line 162: nit: maybe we should make this function return `None` in case of passing items with no generics (use, mod etc.). Just to make a clear distinction between item not having any generics and item that doesn't have any generics by design

src/lang/db/semantic.rs
```diff
@@ -612,3 +620,40 @@ fn find_generated_nodes<'db>(
...
+                db.extern_function_declaration_generic_params(extern_function_id)
+            }
+            _ => return vec![],
+        },
+        LookupItemId::TraitItem(trait_item_id) => {
+            db.trait_generic_params(trait_item_id.trait_id(db))
+        }
+        LookupItemId::ImplItem(impl_item_id) => {
+            db.impl_def_generic_params(impl_item_id.impl_def_id(db))
+        }
```
wawel37, line 654: Correct me if im wrong, but do we want to look for impl/trait generic params, when the lookupitem points to impl/trait **item**, like impl_func, trait_type etc.? Can't they have their own generic parameters?

integraledelebesgue: Of course, they can. We want to look at the whole stack of lookup items above the analyzed identifier because generics can be defined on different levels. Example:

```cairo
trait Trait<T> {
    fn foo<U>(x: T, y: U) -> (T, U) { (x, y) }
}
```

Depending on what parameter we resolve, we need to look at the different level of lookup item - `U` is introduced by the function (a trait item), while `T` is defined by the trait itself.

integraledelebesgue: Of course they can. I overlooked it :)

src/lang/db/semantic.rs
integraledelebesgue: I don't think such distinction is important. We usually know for what item we're obtaining generics for and usually we don't do so for modules or imports

src/lang/db/semantic.rs
integraledelebesgue: Fixed, now item-defined generics are properly resolved

src/lang/db/semantic.rs
```diff
@@ -612,3 +620,48 @@ fn find_generated_nodes<'db>(
...
+            TraitItemId::Type(trait_type_id) => db.trait_type_generic_params(trait_type_id),
+            _ => return vec![],
+        },
+        LookupItemId::ImplItem(impl_item_id) => match impl_item_id {
+            ImplItemId::Function(impl_function_id) => {
+                db.impl_function_generic_params(impl_function_id)
+            }
+            ImplItemId::Type(impl_type_def_id) => db.impl_type_def_generic_params(impl_type_def_id),
+            _ => return vec![],
+        },
```
wawel37, line 662: how about nested impls and traits? I mean, the `LookupItemId::TraitItem::Impl` and same thing for impls. Or is it completely something else?

integraledelebesgue: Traits and impls cannot be nested. The `LookupItemId::TraitItem::Impl` is an associated impl which cannot define its own generics. The same with associated consts and types - they do not introduce new params

wawel37 (approved):

PR #877, opened 2025-08-06
Add a function for formatting traits

integraledelebesgue:
## Stack
* https://github.com/software-mansion/cairols/pull/895
* https://github.com/software-mansion/cairols/pull/878
* https://github.com/software-mansion/cairols/pull/877 ◀️
* https://github.com/software-mansion/cairols/pull/881 
* https://github.com/software-mansion/cairols/pull/876

### DIFF
--- a/src/ide/format/traits.rs
+++ b/src/ide/format/traits.rs
@@ -1,0 +1,40 @@
+use anyhow::Context;
+use cairo_lang_defs::ids::{ImportableId, TraitId};
+use cairo_lang_semantic::db::SemanticGroup;
+use cairo_lang_utils::ordered_hash_map::OrderedHashMap;
+
+use super::types::format_generic_args;
+use crate::lang::db::AnalysisDatabase;
+
+#[allow(dead_code)] // Used later in the stack
+/// Returns a textual representation of a trait with the given [`TraitId`],
+/// consisting of name, path and generic parameters.
+/// Precedes the name with a shortest path allowed by `importables`.
+///
+/// # Example
+/// For `core::num` importable in scope, `Zero` trait is going to be formatted: `num::traits::Zero<T>`.
+pub fn format_trait_path<'db>(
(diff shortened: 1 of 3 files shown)


src/ide/trt.rs
```diff
@@ -0,0 +1,35 @@
+use std::borrow::Cow;
+
```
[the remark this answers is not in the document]
integraledelebesgue: I can rename the short-named modules (`ty.rs`, `trt.rs`) in a separate PR. They can be called, for example, `types.rs` and `traits.rs`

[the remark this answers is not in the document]
integraledelebesgue: Done: https://github.com/software-mansion/cairols/pull/881

src/ide/format/traits.rs
```diff
@@ -0,0 +1,36 @@
...
+    db: &'db AnalysisDatabase,
+    trait_id: TraitId<'db>,
+    importables: &OrderedHashMap<ImportableId<'db>, String>,
+) -> String {
+    let importable = ImportableId::Trait(trait_id);
+
+    let path: Cow<str> = importables
+        .get(&importable)
+        .map(Into::into)
+        .unwrap_or_else(|| Cow::Owned(trait_id.name(db).to_string()));
```
piotmag769, line 24: Is there a case where it may happen?

integraledelebesgue: Probably not. The compiler would have to have inconsistent semantic information

wawel37 (approved):

src/ide/format/traits.rs
integraledelebesgue: It's also going to fail if the given trait is not in scope. I think we can leave it as it is

integraledelebesgue: Done, I added an error with unwrap. We will see if it causes any anexpected issues in the futute

PR #892, opened 2025-08-08
Remove old linter plugin and use query group from linter (#845)

wawel37:


integraledelebesgue (approved):
I assume it's a cherry-pick and doesn't require in-depth review

PR #935, opened 2025-08-20
Fix respecting lint config

piotmag769:


Arcticae (commented):
I get the point of this PR but a q comes to my mind:
Why there's a hashmap and what does it represent in those cases?
Don't we want to merge those potentially? (also applicable for arrays)

I wonder if we should consider using a lib for this so we don't have to consider any more edge cases

piotmag769:
> Why there's a hashmap and what does it represent in those cases?
```rs
pub type CairoLintToolMetadata = OrderedHashMap<String, bool>;
```
Describing lint being off/on.

I believe it was done bcs it was easier to use:
https://github.com/software-mansion/cairo-lint/blob/6bafd85da25ea598f43fc410f1726d8f48a5442f/src/lang/mod.rs#L124
You would have to use `macro_rules` with `stringify!` otherwise and the config would have to be autogenerated from all possible lint kind - there is a task for autogeneration I believe, but not high prio.

integraledelebesgue (approved):

Issue #925, opened 2025-08-20
Use snippets with args names in method/function completions


piotmag769:
E.g. when proposing `func` of signature
```cairo
fn func(a: u8, b: u32) {}
```
it should insert snippets with placeholders containing arg names (check https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax) i.e. `func(a, b)`

PR #939, opened 2025-08-25
Use snippets in dot method completions

piotmag769:
Related to #925

---

**Stack**:
- #942
- #939 ⬅

### DIFF
--- a/src/ide/completion/helpers/snippets.rs
+++ b/src/ide/completion/helpers/snippets.rs
@@ -1,0 +1,20 @@
+use cairo_lang_semantic::Signature;
+use itertools::Itertools;
+
+/// Creates an LSP snippet for the function with the given name the signature.
+///
+/// Example: for a function with a signature
+/// ```cairo
+/// fn xyz(a: u8, b: ByteArray) -> felt252 {}
+/// ```
+/// returns a string "xyz({$1:a}, {$2:b})".
+pub fn snippet_for_function_call(function_name: &str, signature: Signature) -> String {
+    let params_snippet = signature
+        .params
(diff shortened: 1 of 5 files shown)


integraledelebesgue (approved):

wawel37 (commented):
As I've checked the RA, it puts the cursor at the end of the func invocation `.into()<caret>` instead of ours `into(<caret>)` when it comes to funcs with no args. Maybe we should also support that?

piotmag769:
>As I've checked the RA, it puts the cursor at the end of the func invocation .into()<caret> instead of ours into(<caret>) when it comes to funcs with no args. Maybe we should also support that?

After this PR it is exactly how it will behave since no args -> no `$i` in snippet -> cursor at the end of the completion

wawel37 (approved):

src/ide/completion/helpers/snippets.rs
```diff
@@ -0,0 +1,20 @@
...
+use itertools::Itertools;
+
+/// Creates an LSP snippet for the function with the given name the signature.
+///
+/// Example: for a function with a signature
+/// ```cairo
+/// fn xyz(a: u8, b: ByteArray) -> felt252 {}
+/// ```
+/// returns a string "xyz({$1:a}, {$2:b})".
+pub fn snippet_for_function(function_name: &str, signature: Signature) -> String {
```
Draggu, line 11:
```suggestion
pub fn snippet_for_function_call(function_name: &str, signature: Signature) -> String {
```

Draggu (approved):

PR #943, opened 2025-08-26
Make cfg plugin run first

piotmag769:


Arcticae (approved):

src/lang/db/mod.rs
```diff
@@ -147,17 +149,22 @@ impl AnalysisDatabase {
...
     /// Adds proc macro plugin suite to the database.
     ///
-    /// It *prepends* the plugins from the proc macro plugin suite to appropriate
-    /// salsa inputs. It is done to make sure proc macros are resolved first, just like in
+    /// It *prepends* (with the exception of macro plugins, see the code below) the plugins from
+    /// the proc macro plugin suite to appropriate salsa inputs.
+    /// It is done to make sure proc macros are resolved first, just like in
     /// [`crate::project::Crate::apply`].
     pub fn add_proc_macro_plugin_suite(&mut self, crate_input: CrateInput, plugins: PluginSuite) {
         self.with_plugins_mut(
```
wawel37, line 157: nit: maybe we should make the assertion inside the `with_plugins_mut`, so that when we call `set_macro_plugin_overrides_input`, we make sure the first one is the the configplugin?

wawel37 (approved):

src/lang/db/mod.rs
piotmag769: Good idea! Will change it

Arcticae (approved):

Issue #940, opened 2025-08-25
`get_code_actions_for_diagnostics` doesn't work for macro-expanded code


Arcticae:
### Problem

It needs an implementation of resultants usage

### Scarb Version

```text

```

Issue #932, opened 2025-08-20
Suggest quick fixes for typos


piotmag769:
If someone mistypes a name slightly e.g. `self.wirte()` instead of `self.write()` we could suggest a quick fix like "Did you mean write?".
This should be done for:
- [x] Methods #944
- [x] Struct members #956

PR #944, opened 2025-08-26
Add `suggest_similar_method` code action

Arcticae:
With reservation to the macro expanded code, which doesn't work as described in #940

### DIFF
--- a/src/ide/code_actions/mod.rs
+++ b/src/ide/code_actions/mod.rs
@@ -12,6 +12,7 @@ use crate::lang::analysis_context::AnalysisContext;
 use crate::lang::db::{AnalysisDatabase, LsSyntaxGroup};
 use crate::lang::lsp::{LsProtoGroup, ToCairo};
 use crate::project::ConfigsRegistry;
+use std::ops::Not;
 
 mod add_missing_trait;
 mod cairo_lint;
@@ -21,6 +22,7 @@ mod fill_struct_fields;
 mod fill_trait_members;
 mod missing_import;
 mod rename_unused_variable;
+mod suggest_similar_method;
 
 /// Compute commands for a given text document and range. These commands are typically code fixes to
 /// either fix problems or to beautify/refactor code.
@@ -88,7 +90,15 @@ fn get_code_actions_for_diagnostics(
             )
             .to_vec(),
             Some("E0002") => {
-                add_missing_trait::add_missing_trait(db, &ctx, uri.clone()).unwrap_or_default()
+                let fixes = add_missing_trait::add_missing_trait(db, &ctx, uri.clone());
+                if let Some(fixes) = fixes
+                    && fixes.is_empty().not()
+                {
+                    fixes
+                } else {
+                    suggest_similar_method::suggest_similar_method(db, &ctx, uri.clone())
+                        .unwrap_or_default()
+                }
             }
             Some("E0003") => fill_struct_fields::fill_struct_fields(db, ctx.node, params).to_vec(),
             Some("E0004") => fill_trait_members::fill_trait_members(db, &ctx, params).to_vec(),
--- a/src/ide/code_actions/suggest_similar_method.rs
+++ b/src/ide/code_actions/suggest_similar_method.rs
@@ -1,0 +1,89 @@
+use crate::lang::analysis_context::AnalysisContext;
+use crate::lang::db::AnalysisDatabase;
+use crate::lang::importer::import_edit_for_trait_if_needed;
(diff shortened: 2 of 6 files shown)


src/ide/code_actions/suggest_similar_method.rs
```diff
@@ -0,0 +1,82 @@
...
+        .inspect_err(|err| eprintln!("{:?}", err))
+        .ok()?;
+
+    let semantic_db: &dyn SemanticGroup = db.upcast();
+    let ty = semantic_db.expr_semantic(function_with_body, expr_id).ty();
+
+    if ty.is_missing(db) {
+        return None;
+    }
+    eprintln!("ty: {:?}", ty);
```
wawel37, line 37: leftover?

src/ide/code_actions/suggest_similar_method.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    let matching_methods: Vec<String> = method_names
+        .into_iter()
+        .filter_map(|function_id| {
+            let fn_name = function_id.long(db).name(db);
+            let name_matches = text_matches(fn_name, ctx.node.get_text(db));
+            if name_matches {
+                return Some(fn_name);
+            }
+            None
```
wawel37, line 49:
```suggestion
            text_matches(fn_name, ctx.node.get_text(db)).then_some(fn_name)
```

src/ide/code_actions/suggest_similar_method.rs
```diff
@@ -0,0 +1,82 @@
...
+                range: Range {
+                    start: bad_method_name_span.start.to_lsp(),
+                    end: bad_method_name_span.end.to_lsp(),
+                },
+                new_text: method_name.to_string(),
+            };
+            CodeAction {
+                title: format!("Use {method_name} instead"),
+                kind: Some(CodeActionKind::QUICKFIX),
+                is_preferred: Some(only_one_method),
```
wawel37, line 70: maybe we should make the most matching method (float returned from `jaro_winkler`) preferred in case of multiple found?

src/ide/code_actions/suggest_similar_method.rs
Arcticae: functional propaganda

src/ide/code_actions/suggest_similar_method.rs
Arcticae: we can do that 👍

tests/e2e/code_actions/similar_method.rs
```diff
@@ -0,0 +1,150 @@
...
+    mod nested {
+        fn main() {
+            let x = 5_felt252;
+            x.some_me<caret>thood();
+        }
+    }
+    ", @r#"
+    Title: Use some_method instead
+    Add new text: "some_method"
+    At: Range { start: Position { line: 10, character: 10 }, end: Position { line: 10, character: 22 } }
```
piotmag769, line 78: This does not fix it - we should add a proper `use` statement as well here

src/ide/code_actions/mod.rs
```diff
@@ -88,7 +90,14 @@ fn get_code_actions_for_diagnostics(
             )
             .to_vec(),
             Some("E0002") => {
```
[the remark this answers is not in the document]
Arcticae: 1. I will submit a follow up for **struct members**  - added to the issue (#932), unliked from PR

but i can't think of any other case i would want this kind of quick fix 

Do we want to try to find traits/structs which are similar? 
i.e. on `E0006: Identifier not found`
```
struct MyStruct {
   // ...
}
trait MyTraitXd { 
    fn mayFun() -> u32;
}

fn main() {
    MayStruct {}; // <- Did you mean "MyStruct"?
    MayTraitXd {}; // <- Did you mean "MayTrait"?
}
```

RA has this:

image

src/ide/code_actions/mod.rs
```diff
@@ -88,7 +90,14 @@ fn get_code_actions_for_diagnostics(
...
-                add_missing_trait::add_missing_trait(db, &ctx, uri.clone()).unwrap_or_default()
+                let fixes = add_missing_trait::add_missing_trait(db, &ctx, uri.clone());
+                if let Some(fixes) = fixes
+                    && fixes.is_empty().not()
+                {
+                    return fixes;
+                }
+                suggest_similar_method::suggest_similar_method(db, &ctx, uri.clone())
+                    .unwrap_or_default()
```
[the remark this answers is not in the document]
Arcticae: This is a matter of taste

PR #962, opened 2025-09-02
Fix use statements completions

piotmag769:
No longer propose items that are not currently visible.
While it makes sense to do that for completions in a function body since additional use statement can be added, it makes no sense to add another use statement for a use statement completion.
Checked and this is how Intellij does it.


piotmag769:
>Maybe we should also add a test for use my_mod::{<caret>}?

Already tested

PR #965, opened 2025-09-05
Basic support for declarative macro goto

Arcticae:

PR #967, opened 2025-09-05
Expand declarative macro goto definition tests

Arcticae:
**Stack**:
- #974
- #973
- #972
- #967 ⬅
- #965


piotmag769 (approved):
Tests themselves are fine

wawel37 (commented):
Please add a test case with attribute proc macro, as it's done in the same file with the `inline_macro_with_macros` test.

PR #971, opened 2025-09-09
Top level inline macro completions

### LINKED ISSUE #931: Completions for top level inline macros
E.g. `component!`

piotmag769:
Closes #931

### DIFF
--- a/src/ide/completion/mod.rs
+++ b/src/ide/completion/mod.rs
@@ -10,7 +10,6 @@ use cairo_lang_syntax::node::ast;
 use cairo_lang_syntax::node::kind::SyntaxKind;
 use cairo_lang_syntax::node::{SyntaxNode, TypedSyntaxNode};
 use cairo_lang_utils::ordered_hash_set::OrderedHashSet;
-use expr::macro_call::macro_call_completions;
 use function::params::params_completions;
 use function::variables::variables_completions;
 use lsp_types::{CompletionParams, CompletionResponse, CompletionTriggerKind};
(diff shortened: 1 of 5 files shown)


src/ide/completion/mod.rs
```diff
@@ -51,12 +52,7 @@ pub fn complete(params: CompletionParams, db: &AnalysisDatabase) -> Option<Compl
     let was_node_corrected = base_position_node == node;
 
     // There is no completions for these.
-    if matches!(
-        node.kind(db),
-        SyntaxKind::TokenSkipped
-            | SyntaxKind::TriviumSkippedNode
```
piotmag769, line 57: Was never hittable since btw since it is not a token

src/ide/completion/mod.rs
```diff
@@ -71,6 +67,7 @@ pub fn complete(params: CompletionParams, db: &AnalysisDatabase) -> Option<Compl
     // In case we are on eof go back to last non-trivia non-missing node.
     if node.kind(db) == SyntaxKind::SyntaxFile
         || node.ancestor_of_kind(db, SyntaxKind::TerminalEndOfFile).is_some()
+            && node.ancestor_of_kind(db, SyntaxKind::TriviumSkippedNode).is_none()
```
piotmag769, line 70: This makes other completions consistent (as in they work at the end of the file if they did in the middle of them, even when they are not supposed to, but that's not in the scope of this PR) as well

src/ide/completion/path.rs
```diff
@@ -29,7 +29,8 @@ pub fn path_suffix_completions<'db>(
     ctx: &AnalysisContext<'db>,
     was_node_corrected: bool,
 ) -> Vec<CompletionItem> {
-    let (importables, typed_text) = if ctx.node.ancestor_of_kind(db, SyntaxKind::Attribute).is_none()
+    let (importables, typed_text) = if ctx.node.ancestor_of_kind(db, SyntaxKind::TriviumSkippedNode).is_none()
```
piotmag769, line 32: This fixes weird completions like proposing function calls here:
```cairo
mod x {
   functi<caret>
}
```

Draggu (commented):
Add tests for partially typed macros like
```
compil<caret> ();
```

PR #991, opened 2025-09-16
Remove `cairo-lang-proc-macros` and use `dyn Database`

Draggu:


src/lang/db/syntax.rs
piotmag769, line 1: Add `#[salsa::tracked]` everywhere

src/ide/code_actions/cairo_lint.rs
```diff
@@ -30,10 +30,10 @@ pub fn cairo_lint<'db>(
     // We collect the semantic diagnostics, as the unused imports diagnostics (which come from the semantic diags),
     // can be fixed with the linter.
     let semantic_diags = db.module_semantic_diagnostics(module_id).ok()?;
-    let linter_diags = db.linter_diagnostics(linter_params, module_id).into_iter().map(|diag| {
+    let linter_diags = db.linter_diagnostics(linter_params, module_id).iter().map(|diag| {
         SemanticDiagnostic::new(
             StableLocation::new(diag.stable_ptr),
-            SemanticDiagnosticKind::PluginDiagnostic(diag),
+            SemanticDiagnosticKind::PluginDiagnostic(diag.clone()),
```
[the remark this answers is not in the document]
Draggu: Lint changed and returns refs, so we clone on demand here now

src/lang/db/semantic.rs
piotmag769, line 1: Add `#[salsa::tracked]` everywhere

src/lang/lsp/ls_proto_group_test.rs
```diff
@@ -32,9 +32,9 @@ fn file_url() {
     check("file:///", FileLongId::OnDisk("/".into()));
 
     // NOTE: We expect that Salsa is assigning sequential numeric ids to files,
-    //   hence numbers 10242 and 10243 appear further down.
+    //   hence numbers 8194 and 8195 appear further down.
```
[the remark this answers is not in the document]
Draggu: These numbers change with every salsa/compiler change. I would prefer to rework this test NOT to assert these numbers at all

src/lang/proc_macros/db.rs
```diff
@@ -1,93 +1,109 @@
...
 fn get_stored_derive_expansion(
-    db: &dyn ProcMacroGroup,
+    db: &dyn Database,
     params: PlainExpandDeriveParams,
 ) -> Option<ProcMacroResult> {
-    db.derive_macro_resolution().get(&params).cloned()
+    db.proc_macro_input().derive_macro_resolution(db).get(&params).cloned()
 }
 
+// #[salsa::tracked]
```
piotmag769, line 89: Ye, prob not worth to have it as a query since the lookup should be cheap, please remove comments

src/lang/db/mod.rs
```diff
@@ -58,10 +57,6 @@ impl AnalysisDatabase {
         init_defs_group(&mut db);
         init_semantic_group(&mut db);
         init_lowering_group(&mut db, InliningStrategy::Default, None);
-        // proc-macro-server can be restarted many times but we want to keep these data across
-        // multiple server starts, so init it once per database, not per server.
-        init_proc_macro_group(&mut db);
```
piotmag769, line 63: Shouldn't we set all resolutions to default? It will prob panic on `get_*` methods if someone tries to do that without setting anything first

src/lang/proc_macros/client/plain_request_response.rs
```diff
@@ -1,25 +1,26 @@
+use salsa::Update;
 use scarb_proc_macro_server_types::methods::expand::{
     ExpandAttributeParams, ExpandDeriveParams, ExpandInlineMacroParams,
 };
 use scarb_proc_macro_server_types::scope::ProcMacroScope;
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Update)]
```
piotmag769, line 8: [nit] I think full path i. e. `salsa::Update` is clearer

Draggu: Disagree, this is consistent because we don't use e.g. `serde::Serialize`

src/lang/db/mod.rs
Draggu: It won't input access is a query without params, it sets defaults on the first call. It panics in the compiler because they use `Option` that is later `.unwrap()`, and this panics.

wawel37 (approved):

piotmag769 (approved):

PR #998, opened 2025-09-18
Use nightly Rust toolchain for the format check

integraledelebesgue:
## Rationale
I would like to enforce a strict format of imports.


wawel37 (approved):
If those unstable features in rust one day become stable, we can just use the nightly for the `fmt` only, right?

piotmag769 (approved):

Draggu (approved):

Issue #924, opened 2025-08-20
Show types/signatures for completions


piotmag769:
Function signatures for method/function completions, types for members/variables etc.

PR #999, opened 2025-09-19
Add signature for functions in dot completions

Arcticae:
Related: #924

### DIFF
--- a/src/ide/completion/dot_completions.rs
+++ b/src/ide/completion/dot_completions.rs
@@ -1,4 +1,4 @@
-use cairo_lang_defs::ids::{NamedLanguageElementId, TopLevelLanguageElementId, TraitFunctionId};
+use cairo_lang_defs::ids::{NamedLanguageElementId, TraitFunctionId};
 use cairo_lang_filesystem::ids::CrateId;
 use cairo_lang_semantic::db::SemanticGroup;
 use cairo_lang_semantic::items::function_with_body::{
@@ -18,6 +18,7 @@ use lsp_types::{CompletionItem, CompletionItemKind, InsertTextFormat};
 use tracing::debug;
 
 use crate::ide::completion::helpers::binary_expr::dot_rhs::dot_expr_rhs;
+use crate::ide::completion::helpers::formatting::generate_abbreviated_signature;
 use crate::ide::completion::helpers::snippets::snippet_for_function_call;
 use crate::ide::completion::{CompletionItemOrderable, CompletionRelevance};
 use crate::ide::format::types::format_type;
@@ -124,8 +125,9 @@ fn completion_for_method<'db>(
     let name = trait_function.name(db).to_string(db);
     let signature = db.trait_function_signature(trait_function).ok()?;
 
-    // TODO(spapini): Add signature.
-    let detail = trait_id.full_path(db);
+    let abbreviated_signature =
+        generate_abbreviated_signature(db, signature, Some(trait_function.trait_id(db)));
+
     let mut additional_text_edits = vec![];
 
     // If the trait is not in scope, add a use statement.
@@ -138,7 +140,7 @@ fn completion_for_method<'db>(
             label: format!("{name}()"),
             insert_text: Some(snippet_for_function_call(db, &name, signature)),
             insert_text_format: Some(InsertTextFormat::SNIPPET),
-            detail: Some(detail),
+            detail: Some(abbreviated_signature),
             kind: Some(CompletionItemKind::METHOD),
             additional_text_edits: Some(additional_text_edits),
             ..CompletionItem::default()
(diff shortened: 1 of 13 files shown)


src/ide/completion/dot_completions.rs
```diff
@@ -124,8 +124,9 @@ fn completion_for_method<'db>(
     let name = trait_function.name(db);
     let signature = db.trait_function_signature(trait_function).ok()?;
 
-    // TODO(spapini): Add signature.
-    let detail = trait_id.full_path(db);
+    let explicit_signature = signature.stable_ptr.0.lookup(db).get_text_without_trivia(db);
```
Draggu, line 127: There still can be inner trivia. Use method from doc that allows obtaining it without inner trivia

src/ide/completion/dot_completions.rs
```diff
@@ -148,3 +149,15 @@ fn completion_for_method<'db>(
     };
     Some(completion)
 }
+
+// Delete all the whitespace and newlines from the signature, format it and add `fn` marker.
```
Draggu, line 153: Most of this logic will be unnecessary if the correct text obtaining method is used, also this will break const generics if used with `'felt   literal'`.

Draggu (changes requested):
Add test with ```
param: ConstGenericType<'a   a'>
```
where generic is felt with multiple consecutive whitespaces

File: src/ide/completion/dot_completions.rs (hunk 1 above)
Comment (wawel37) [reply src/ide/completion/dot_completions.rs:127]: Check https://github.com/starkware-libs/cairo/blob/8ee945ce7d23eebdad149c4646fe43a2a6712a79/crates/cairo-lang-syntax/src/node/mod.rs#L221, maybe will be helpful

**Review by wawel37 — APPROVED** [review]

File: tests/e2e/completions/dot.rs [hunk 3]
```diff
@@ -234,12 +234,12 @@ fn with_deref_starknet() {
...
     [[completions]]
     completion_label = "destruct()"
-    detail = "core::traits::Destruct"
+    detail = "fn(self: T) -> () nopanic"
     insert_text = "destruct()"
 
     [[completions]]
     completion_label = "emit()"
-    detail = "core::starknet::event::EventEmitter"
+    detail = "fn(ref self: T, event: S) -> ()"
```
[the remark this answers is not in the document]
Arcticae: idk about this - i went back and forth with this, and left it this way since intelij shows the tuple type. Having it side by side with other function makes more sense - for consistency (visually)

src/ide/completion/helpers/formatting.rs
```diff
@@ -0,0 +1,55 @@
...
+        if !first {
+            s.push_str(", ");
+        }
+        first = false;
+
+        let modifier = get_relevant_modifier(&param.mutability);
+        if !modifier.is_empty() {
+            s.push_str(modifier);
+            s.push(' ');
+        }
```
Draggu, line 35: param mutability modifier is an internal implementation detail, it should not be included in its public interface

src/ide/format/types.rs
```diff
@@ -40,23 +41,36 @@ pub fn format_type<'db>(
...
             format!(
                 "fn ({}) -> {}",
-                closure.param_tys.iter().map(|ty| format_type(db, *ty, importables)).join(", "),
-                format_type(db, closure.ret_ty, importables)
+                closure
+                    .param_tys
+                    .iter()
+                    .map(|ty| format_type(db, *ty, importables, trait_id))
+                    .join(", "),
+                format_type(db, closure.ret_ty, importables, trait_id)
```
Draggu, line 63: Can we make this logic and `generate_abbreviated_signature` common?

src/ide/completion/helpers/formatting.rs
Arcticae: i skipped mut

src/ide/completion/helpers/formatting.rs
```diff
@@ -0,0 +1,60 @@
...
+    let mut s = String::from("fn(");
+
+    let formatted_params = signature
+        .params
+        .iter()
+        .map(|param| {
+            let mut s = String::new();
+            let modifier = get_relevant_modifier(&param.mutability);
+
+            if !modifier.is_empty() && modifier != "mut" {
```
Draggu, line 34: match on `param.mutability` instead of string comparision

src/ide/completion/helpers/formatting.rs
```diff
@@ -0,0 +1,60 @@
...
+        .join(", ");
+
+    s.push_str(&formatted_params);
+    s.push_str(") -> ");
+
+    let ret_text = format_type(db, signature.return_type, &importables, trait_id);
+
+    s.push_str(&ret_text);
+
+    if !signature.panicable {
```
Draggu, line 55: can we add `extern` before `fn` too?

Arcticae: unnecessary noise

Draggu (approved):

PR #1002, opened 2025-09-22
Struct initialization completion

wawel37:


src/ide/completion/path.rs
```diff
@@ -256,3 +265,47 @@ pub fn path_prefix_completions<'db>(
...
+    ctx: &AnalysisContext<'db>,
+    importable: &ImportableId<'db>,
+) -> Option<String> {
+    let statement_let = &ctx.node.ancestor_of_type::<StatementLet>(db);
+    let statement_expr = &ctx.node.ancestor_of_type::<StatementExpr>(db);
+    let importable_node = importable_syntax_node(db, *importable)?;
+
+    if (statement_let.is_some() || statement_expr.is_some())
+        && let Some(_path) = &ctx.node.ancestor_of_type::<ExprPath>(db)
+        && importable_node.kind(db) == SyntaxKind::ItemStruct
```
Draggu, line 280: Should be matched on importable itself

src/ide/completion/path.rs
```diff
@@ -108,9 +111,15 @@ pub fn path_suffix_completions<'db>(
             let is_current_crate = importable_crate == current_crate;
             let is_core = *importable_crate.long(db) == CrateLongId::core(db);
 
+            let struct_initialization_text =
```
Draggu, line 114: It is not extendible for other items than structs, extract importable match outside `struct_initialization_completion_text`

tests/e2e/completions/path.rs
```diff
@@ -27,9 +27,11 @@ fn single_element_path() {
 
     [[completions]]
     completion_label = "ByteA_ActuallyNotByteArray"
+    insert_text = "ByteA_ActuallyNotByteArray {}"
 
     [[completions]]
     completion_label = "ByteArray"
+    insert_text = "ByteArray { data: $1, pending_word: $2, pending_word_len: $3 }"
```
Draggu, line 34: `ByteArray` fields are private, so it is NOT constructible. We should not propose its struct constructor.
See https://github.com/software-mansion/cairols/blob/19409388ea8b18eeecd09ca94ca7a01c88e2581d/src/ide/completion/struct_constructor.rs#L74-L79

src/ide/completion/path.rs
```diff
@@ -256,3 +265,47 @@ pub fn path_prefix_completions<'db>(
...
+                    member.name(db).as_syntax_node().get_text_without_trivia(db).long(db).as_str(),
+                    index + 1,
+                )
+            })
+            .join(", ");
+
+        return Some(if args.is_empty() {
+            format!("{} {{}}", struct_name)
+        } else {
+            formatdoc!(
```
Draggu, line 301: Why do we need `formatdoc!` here? `format!` is enough

src/ide/completion/path.rs
```diff
@@ -256,3 +265,47 @@ pub fn path_prefix_completions<'db>(
...
+            struct_node.name(db).as_syntax_node().get_text_without_trivia(db).long(db).as_str();
+        let args = struct_node
+            .members(db)
+            .elements(db)
+            .enumerate()
+            .map(|(index, member)| {
+                format!(
+                    "{}: ${}",
+                    member.name(db).as_syntax_node().get_text_without_trivia(db).long(db).as_str(),
+                    index + 1,
```
Draggu, line 293: Can we start from 0? If so, please do it this way, it is simpler; if not, please adda  comment explaining why

src/ide/completion/path.rs
```diff
@@ -256,3 +265,47 @@ pub fn path_prefix_completions<'db>(
...
+fn struct_initialization_completion_text<'db>(
+    db: &'db AnalysisDatabase,
+    ctx: &AnalysisContext<'db>,
+    importable: &ImportableId<'db>,
+) -> Option<String> {
+    let statement_let = &ctx.node.ancestor_of_type::<StatementLet>(db);
+    let statement_expr = &ctx.node.ancestor_of_type::<StatementExpr>(db);
+    let importable_node = importable_syntax_node(db, *importable)?;
+
+    if (statement_let.is_some() || statement_expr.is_some())
```
Draggu, line 278: This may not be enough for the closure parameter type inside another expression

```
let a = {
    |b: some::module::StructA<caret>|
}
```

Will trigger this while it should not

integraledelebesgue (approved):

PR #1011, opened 2025-09-30
Completions for declarative macros

wawel37:
Need to wait for https://github.com/starkware-libs/cairo/pull/8488 to be merged

### DIFF
--- a/src/ide/completion/path.rs
+++ b/src/ide/completion/path.rs
@@ -1,4 +1,4 @@
-use cairo_lang_defs::ids::{LanguageElementId, NamedLanguageElementId};
+use cairo_lang_defs::ids::{ImportableId, LanguageElementId, NamedLanguageElementId};
 use cairo_lang_filesystem::ids::{CrateLongId, SmolStrId};
 use cairo_lang_semantic::diagnostic::{NotFoundItemType, SemanticDiagnostics};
 use cairo_lang_semantic::items::enm::EnumSemantic;
@@ -67,6 +67,15 @@ pub fn path_suffix_completions<'db>(
 
     let current_crate = ctx.module_file_id.0.owning_crate(db);
 
+    db.visible_importables_from_module(ctx.module_file_id)
+        .unwrap()
+        .iter()
+        .filter(|importable| importable.1.contains("haslo"))
+        .for_each(|importable| {
+            eprintln!("Ignoring macro importable: {:?}", importable.1);
+        });
+
+    eprintln!("=======");
     let mut completions: Vec<CompletionItemOrderable> = importables
         .iter()
         .filter_map(|(importable, path_str)| {
@@ -105,13 +114,20 @@ pub fn path_suffix_completions<'db>(
             } else {
                 None
             };
+
+            let is_declarative_macro = matches!(importable, ImportableId::MacroDeclaration(_));
+
             let importable_crate = importable_crate_id(db, *importable);
             let is_current_crate = importable_crate == current_crate;
             let is_core = *importable_crate.long(db) == CrateLongId::core(db);
 
             Some(CompletionItemOrderable {
                 item: CompletionItem {
-                    label: last_segment.to_string(),
+                    label: if is_declarative_macro {
+                        format!("{}!", last_segment.to_string())
+                    } else {
+                        last_segment.to_string()
+                    },
                     kind: Some(importable_completion_kind(*importable)),
                     label_details: Some(CompletionItemLabelDetails {
                         detail: None,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -244,7 +244,6 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 [[package]]
 name = "cairo-lang-casm"
 version = "2.12.3"
-source = "git+https://github.com/starkware-libs/cairo?rev=2b3ba4e10df6555698c6e03d0aae931e44d3be2f#2b3ba4e10df6555698c6e03d0aae931e44d3be2f"
 dependencies = [
  "cairo-lang-utils",
  "indoc",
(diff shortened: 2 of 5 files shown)


integraledelebesgue (approved):

Arcticae (approved):

src/ide/completion/path.rs
```diff
@@ -125,12 +137,25 @@ pub fn path_suffix_completions<'db>(
                 _ => None,
             };
 
+            let snippet_text = get_snippet_text(
```
Arcticae, line 140: why is it called twice (here and below)?

src/ide/completion/path.rs
```diff
@@ -108,11 +119,12 @@ pub fn path_suffix_completions<'db>(
             } else {
                 None
             };
+
             let importable_crate = importable_crate_id(db, *importable);
             let is_current_crate = importable_crate == current_crate;
             let is_core = *importable_crate.long(db) == CrateLongId::core(db);
 
-            let struct_initialization_text = match importable {
+            let struct_initialization_snippet_text = match importable {
```
Arcticae, line 127: Perhaps we should calculate snippet text here instead of moving it around and checking if it's `None`

src/ide/completion/path.rs
```diff
@@ -73,6 +79,11 @@ pub fn path_suffix_completions<'db>(
     let mut completions: Vec<CompletionItemOrderable> = importables
         .iter()
         .filter_map(|(importable, path_str)| {
+            let is_declarative_macro = matches!(importable, ImportableId::MacroDeclaration(_));
```
Arcticae, line 82: Can we move it to the `match` logic below?

src/ide/completion/path.rs
```diff
@@ -125,12 +137,25 @@ pub fn path_suffix_completions<'db>(
...
+                    label: if is_declarative_macro {
+                        format!("{}!", last_segment)
+                    } else {
+                        last_segment.to_string()
+                    },
+                    insert_text: get_snippet_text(
+                        last_segment,
+                        is_declarative_macro,
+                        struct_initialization_snippet_text.clone(),
+                    ),
```
Arcticae, line 157: Those two could be calculated in tandem i think, making for simpler logic to track

src/ide/completion/path.rs
```diff
@@ -34,22 +37,25 @@ pub fn path_suffix_completions<'db>(
     db: &'db AnalysisDatabase,
     ctx: &AnalysisContext<'db>,
     was_node_corrected: bool,
+    only_top_level_declarative_macros: bool,
```
Arcticae, line 40: I'm not a big fan of this function right now, it has not one (which i already disliked before) but two boolean parameters

We should split this up somehow so it's more modular and compose 2 different functions from this (we use it in 2 places with `true` and `false` respectively)

src/ide/completion/path.rs
```diff
@@ -34,22 +37,25 @@ pub fn path_suffix_completions<'db>(
...
+        db,
+        ctx,
+        was_node_corrected,
+        only_top_level_declarative_macros,
+    ) {
+        Some(importables) => importables,
+        None => return Default::default(),
+    };
+
+    let typed_text = match ctx
```
Arcticae, line 52: i'd use `.unwrap_or_else`

src/ide/completion/path.rs
```diff
@@ -34,22 +37,25 @@ pub fn path_suffix_completions<'db>(
...
-        && dot_expr_rhs(db, &ctx.node, was_node_corrected).is_none()
-        // Enum patterns are handled in a separate function.
-        && ctx.node.ancestor_of_kind(db, SyntaxKind::PatternEnum).is_none()
-        && let Some(importables) = db.visible_importables_from_module(ctx.module_file_id)
-        && let Some(typed_text_segments) = ctx
-            .node
-            .ancestor_of_type::<ExprPath>(db)
-            .map(|path| path.segments(db).elements(db).collect_vec())
-        && !typed_text_segments.is_empty()
+    let importables = match get_importables_for_path_suffix_completions(
```
Arcticae, line 42: `.unwrap_or_else`

src/ide/completion/path.rs
```diff
@@ -277,6 +302,37 @@ pub fn path_prefix_completions<'db>(
     })
 }
 
+fn get_importables_for_path_suffix_completions<'db>(
```
Arcticae, line 305: I don't see a benefit from putting this in a function

src/ide/completion/path.rs
```diff
@@ -277,6 +302,37 @@ pub fn path_prefix_completions<'db>(
...
+    db: &'db AnalysisDatabase,
+    ctx: &AnalysisContext<'db>,
+    was_node_corrected: bool,
+    only_top_level_declarative_macros: bool,
+) -> Option<Arc<OrderedHashMap<ImportableId<'db>, String>>> {
+    if (ctx.node.ancestor_of_kind(db, SyntaxKind::TriviumSkippedNode).is_none()
```
Arcticae, line 311: i don't understand this `if` (first condition)

src/ide/completion/path.rs
wawel37: im getting:
```
warning: unneeded `return` statement
  --> src/ide/completion/path.rs:56:28
   |
56 |         .unwrap_or_else(|| return Default::default());
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
   = note: `#[warn(clippy::needless_return)]` on by default
help: remove `return`
   |
56 -         .unwrap_or_else(|| return Default::default());
56 +         .unwrap_or_else(|| Default::default());
```

src/ide/completion/path.rs
wawel37: `ctx.node.ancestor_of_kind(db, SyntaxKind::TriviumSkippedNode).is_none()` is to make sure we are not at the top level, but in case of top level declarative macros completions, we actually want to ignore that check, to let the completion work for those also

src/ide/completion/path.rs
Arcticae: it's basically the same if (only thing the function has) - why not calculate it in the same branch? if not, let's have separate function that does the same thing, or inline it as well. now it's mixed and it seems weird

PR #1034, opened 2025-10-15
Display path for every completion / support declarative macros in completions

wawel37:


src/ide/completion/path.rs
```diff
@@ -1,179 +1,75 @@
...
-use crate::lang::importer::new_import_edit;
-use crate::lang::text_matching::text_matches;
+use crate::lang::db::AnalysisDatabase;
 use crate::lang::visibility::peek_visible_in_with_edition;
 
 /// Treats provided path as suffix, proposing elements that can prefix this path.
 pub fn path_suffix_completions<'db>(
     db: &'db AnalysisDatabase,
-    ctx: &AnalysisContext<'db>,
+    ctx: &'db AnalysisContext<'db>,
```
Draggu: This is incorrect, context is not owned by db

PR #1048, opened 2025-10-22
Order prelude items correctly in completions

wawel37:


src/ide/completion/helpers/item.rs
```diff
@@ -28,8 +28,9 @@ pub enum CompletionRelevance {
     Lowest = 0,
     Low = 1,
     Medium = 2,
-    High = 3,
-    Highest = 4,
+    MediumHigh = 3,
```
Arcticae: Let's make those names actually mean something semantically (associate it with some item types etc. or anything that actually leads us to conclusion where should we place the item in the list).

src/ide/completion/helpers/item.rs
```diff
@@ -153,6 +156,7 @@ impl ImportableCompletionItem<'_> {
             None
         };
 
+        let is_in_scope = additional_text_edits.is_none();
```
Draggu: There is previously `is_not_in_scope` var and now we have `is_in_scope` with different logic. Please change logic/name

PR #1075, opened 2025-11-04
Run heavy tests sequentially

Arcticae:


.config/nextest.toml
```diff
@@ -0,0 +1,4 @@
+[profile.ci]
+[[profile.ci.overrides]]
+filter = 'test(/^(code_lens|completions|find_references|goto_definition|macros|rename)::.*/)'
+threads-required = "num-cpus"
```
Draggu: That's not obvious, please link the doc here

PR #1088, opened 2025-11-14
Fix: Analysis progress on cancel and diagnostics re-run on empty responses

Arcticae:
This was not tracked before, the condition would be checked before the cancellation flag would be set which would cause the condition to wrongly evaluate as end of analysis.

Also adds a cancelation on defined macros response with empty array, which would not cause the cancellation to ocur correctly

### DIFF
--- a/src/lang/proc_macros/controller.rs
+++ b/src/lang/proc_macros/controller.rs
@@ -235,9 +235,19 @@ impl ProcMacroClientController {
             match params {
                 RequestParams::DefinedMacros(params) => {
                     let defined_macros = parse_response::<DefinedMacrosResponse>(response)?;
-                    self.apply_defined_macros_response(db, params.workspace, defined_macros);
+                    self.apply_defined_macros_response(
+                        db,
+                        params.workspace,
+                        defined_macros.clone(),
+                    );
                     self.try_load_proc_macro_cache(db, config);
                     try_request_semantic_tokens_refresh(client_capabilities, requester);
+                    if defined_macros.macros_for_cu_components.is_empty() {
+                        // We have to make sure that we re-run the diagnostics here even if there were no macros defined
+                        // since we need the tick of diagnostics to be triggered even on empty response.
+                        // Analysis won't stop otherwise, because the end condition won't be checked without the tick end.
+                        db.cancel_all();
+                    }
                 }
                 RequestParams::ExpandAttribute(params) => {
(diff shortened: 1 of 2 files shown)


src/lang/proc_macros/controller.rs
```diff
@@ -235,9 +235,17 @@ impl ProcMacroClientController {
             match params {
                 RequestParams::DefinedMacros(params) => {
                     let defined_macros = parse_response::<DefinedMacrosResponse>(response)?;
-                    self.apply_defined_macros_response(db, params.workspace, defined_macros);
+                    self.apply_defined_macros_response(
+                        db,
+                        params.workspace,
+                        defined_macros.clone(),
```
Draggu: Dont' clonse, just check if empty earlier

src/lang/proc_macros/controller.rs
```diff
@@ -235,9 +235,17 @@ impl ProcMacroClientController {
...
+                    );
                     self.try_load_proc_macro_cache(db, config);
                     try_request_semantic_tokens_refresh(client_capabilities, requester);
+                    if defined_macros.macros_for_cu_components.is_empty() {
+                        // We have to make sure that we re-run the diagnostics here even if there were no macros defined
+                        db.cancel_all();
+                    }
```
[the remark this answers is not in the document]
piotmag769: Add a comment that explains why we do this then

src/lang/proc_macros/controller.rs
```diff
@@ -235,9 +235,19 @@ impl ProcMacroClientController {
...
+                        db,
+                        params.workspace,
+                        defined_macros.clone(),
+                    );
                     self.try_load_proc_macro_cache(db, config);
                     try_request_semantic_tokens_refresh(client_capabilities, requester);
+                    if defined_macros.macros_for_cu_components.is_empty() {
+                        // We have to make sure that we re-run the diagnostics here even if there were no macros defined
+                        // since we need the tick of diagnostics to be triggered even on empty response.
+                        // Analysis won't stop otherwise, because the end condition won't be checked without the tick end.
```
piotmag769:
```suggestion
                        // We have to make sure that we re-run the diagnostics here even if there were no macros defined
                        // since we need the diagnostics end tick to be finish analysis.
                        // Analysis won't stop otherwise, because the end condition won't be checked without the tick end.
```

PR #1111, opened 2025-11-24
Fix/scarb metadata no eof

Arcticae:


Arcticae:
Makes other issues worse, closing

PR #1114, opened 2025-11-26
Verify if content of the mapped-back document highlight matches

### LINKED ISSUE #824: `HasComponent` does highlight trait functions bodies in impls within the component
The description is a bit vague but it comes down to this: 

Image

Arcticae:
Closes #824


Draggu (changes requested):

Arcticae:
Discussed offline, takeaways:
This is not-ideal solution to this but the macros are prone to this type of errors made by proc-macro devs
It's a possible solution but we will try to solve it in the compiler instead

PR #1117, opened 2025-11-28
Fix find references for annotated structs

### LINKED ISSUE #819: Support find references for `Storage` struct fields
It works when invoked on a `self.<storage_field>` usage, but on definition (in `Storage` struct def itself) it does not work.

See demo:

https://github.com/user-attachments/assets/8d69a424-27d7-490a-b881-ef04c52bd7eb

THenry14:
Closes #819

### DIFF
--- a/src/lang/usages/mod.rs
+++ b/src/lang/usages/mod.rs
@@ -13,7 +13,7 @@ use memchr::memmem::Finder;
 use search_scope::SearchScope;
 
 use crate::lang::db::{AnalysisDatabase, LsSyntaxGroup};
-use crate::lang::defs::{ResolvedItem, SymbolDef, SymbolSearch};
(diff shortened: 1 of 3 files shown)


src/lang/usages/mod.rs
```diff
@@ -192,9 +192,35 @@ impl<'db> FindUsages<'db> {
...
+            }
+        }
+
+        // We check if a symbol is a Member of a struct, then compare its item_id with a context
+        // of found symbol definitions and declarations
+        let symbol_in_context = match &self.symbol {
+            SymbolDef::Member(member_def) => {
+                let id = member_def.struct_item().lookup_item_id();
+
+                let in_context = |m: &MemberDef| m.struct_item().context_items().contains(&id);
```
Arcticae: This is a bit of a weird heuristic - why would one compare `lookup_item_id` to check if the definition/declaration is equal?

PR #1125, opened 2025-12-03
Make dot method exprs propose completions even with parens after caret

wawel37:


Draggu (changes requested):
This will work only without macros, you pass node from og file and compares it with resultants

src/ide/completion/dot_completions.rs
```diff
@@ -41,6 +42,20 @@ fn dot_completions_ex<'db>(
...
     let expr = dot_expr_rhs(db, &ctx.node, was_node_corrected)?;
+    let rhs = match expr.rhs(db) {
+        Expr::FunctionCall(function_call) => Some(function_call.path(db)),
+        Expr::Path(path) => Some(path),
+        _ => None,
+    }
+    .and_then(|rhs| Some(rhs.segments(db).elements(db).next()?.as_syntax_node()));
+
+    // This way we ignore `my_struct.method(<caret>)` cases, but make sure to allow `my_struct.method(arg1, arg2, my_struct2.method<caret>())` cases.
+    if let Some(rhs) = rhs
```
Draggu: `is_some_and`

PR #1132, opened 2025-12-10
Use commit in proc macro cache file name

Draggu:
**Stack**:
- #1138
- #1135
- #1134
- #1133
- #1132 ⬅

### DIFF
--- a/src/lang/proc_macros/cache.rs
+++ b/src/lang/proc_macros/cache.rs
@@ -14,6 +14,7 @@ use bincode::{
     config::standard,
     serde::{decode_from_slice, encode_to_vec},
 };
+use git_version::git_version;
 use salsa::{Database, Setter};
 use scarb_proc_macro_server_types::methods::ProcMacroResult;
 use serde::{Deserialize, Serialize, ser::SerializeMap};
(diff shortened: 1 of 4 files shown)


src/lang/proc_macros/cache.rs
```diff
@@ -163,11 +164,16 @@ pub fn try_load_proc_macro_cache(db: &mut dyn Database, config: &Config) {
...
         cache_path.push("cairo-language-server");
-        cache_path.push("proc_macro.cache");
+        cache_path.push(cache_file_name());
         cache_path
     })
 }
 
+fn cache_file_name() -> String {
+    // Use commit for file name, so each LS version (even dev ones) will use different cache
+    format!("{}_proc_macro.cache", git_version!())
```
THenry14: this depends on git binary - what happens if it is not present in PATH, or cwd is not under git? maybe we should have some kind of fallback?

Draggu: This is compile time, as long as there is git on our/scarb CI it will work, every dev need git as well but this is guaranteed

Arcticae (approved):

PR #1146, opened 2025-12-16
Proc macro cache use `create_output_dir`

Draggu:


src/lang/proc_macros/cache.rs
```diff
@@ -129,7 +130,9 @@ pub fn save_proc_macro_cache(db: &dyn Database) {
 
     let buffer = encode_to_vec(resolution, standard()).expect("serialize should not fail");
 
-    let _ = fs::create_dir_all(cache_path.parent().expect("parent must exist"));
+    let cache_path_parent = cache_path.parent().expect("LS subdir must exist");
```
maciektr: nit: cache_path that takes target_dir as arg would be more defensive. This is fine too.

integraledelebesgue (approved):

PR #1171, opened 2026-01-15
Implement semantic highlighting for intra-doc links

Arcticae:

PR #1172, opened 2026-01-15
Doc links gotos

Arcticae:
Base: #1171

### DIFF
--- a/src/ide/navigation/goto_definition.rs
+++ b/src/ide/navigation/goto_definition.rs
@@ -4,13 +4,13 @@ use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_filesystem::ids::{
     FileId, FileKind, FileLongId, SmolStrId, SpanInFile, VirtualFile,
 };
-use cairo_lang_filesystem::span::{TextPosition, TextSpan};
+use cairo_lang_filesystem::span::{TextOffset, TextPosition, TextSpan, TextWidth};
 use cairo_lang_parser::db::ParserGroup;
 use cairo_lang_semantic::diagnostic::{NotFoundItemType, SemanticDiagnostics};
 use cairo_lang_semantic::expr::inference::InferenceId;
 use cairo_lang_semantic::lsp_helpers::LspHelpers;
-use cairo_lang_semantic::resolve::{AsSegments, ResolutionContext, Resolver};
-use cairo_lang_syntax::node::ast::TerminalIdentifier;
+use cairo_lang_semantic::resolve::{ResolutionContext, Resolver};
+use cairo_lang_syntax::node::ast::{PathSegment, TerminalIdentifier};
 use cairo_lang_syntax::node::kind::SyntaxKind;
 use cairo_lang_syntax::node::{SyntaxNode, TypedSyntaxNode, ast};
 use cairo_lang_utils::Intern;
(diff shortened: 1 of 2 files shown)


src/ide/navigation/goto_definition.rs
```diff
@@ -36,6 +48,107 @@ pub fn goto_definition(
...
+    db: &AnalysisDatabase,
+    file: FileId<'_>,
+    position: TextPosition,
+) -> Option<Location> {
+    let cursor_offset = position.offset_in_file(db, file)?;
+    let node = db.find_syntax_node_at_position(file, position)?;
+    let token = match node.kind(db) {
+        SyntaxKind::TokenSingleLineDocComment | SyntaxKind::TokenSingleLineInnerComment => node,
+        _ => return None,
+    };
```
integraledelebesgue:
```suggestion
    let token = db.find_syntax_node_at_position(file, position)?;
    if !matches(token.kind(db), SyntaxKind::TokenSingleLineDocComment | SyntaxKind::TokenSingleLineInnerComment) {
        return None;
    };
```

src/ide/navigation/goto_definition.rs
```diff
@@ -36,6 +48,107 @@ pub fn goto_definition(
...
+    let token_span = token.span_without_trivia(db);
+    if !token_span.contains(TextSpan::cursor(cursor_offset)) {
+        return None;
+    }
+
+    let token_text = token.text(db)?.to_string(db);
+    let content = &token_text[3..];
+    // Convert absolute cursor offset to a doc-comment content offset (skip leading /// or //!).
```
integraledelebesgue:
```suggestion
    // Convert absolute cursor offset to an offset relative to the doc-comment (excluding leading /// or //!).
```

src/ide/navigation/goto_definition.rs
```diff
@@ -36,6 +48,107 @@ pub fn goto_definition(
...
+
+    let links = parse_doc_links(content);
+    let link = links.into_iter().find(|link| link.range.contains(&rel_offset))?;
+    if !link.label_range.contains(&rel_offset) {
+        return None;
+    }
+    let label_offset = rel_offset.checked_sub(link.label_range.start)?;
+    let link_prefix = doc_link_prefix(link.label_text.trim(), label_offset)?;
+
+    let segments = parse_doc_link_path(db, file, token_span, link_prefix.trim())?;
```
integraledelebesgue: Passing `token_span` here will make the virtual file created by `parse_doc_link_path` mapped to the whole comment instead of only the link label. Is it somehow relevant?

src/ide/navigation/goto_definition.rs
```diff
@@ -36,6 +48,117 @@ pub fn goto_definition(
...
+    let cursor_span = TextSpan::cursor(cursor);
+    let segments = expr_path.segments(db).elements_vec(db);
+
+    for (idx, segment) in segments.iter().enumerate() {
+        let span = segment.as_syntax_node().span_without_trivia(db);
+        if span.contains(cursor_span) {
+            return Some(segments.into_iter().take(idx + 1).collect());
+        }
+    }
+    None
```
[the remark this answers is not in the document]
Arcticae: I changed it, but preserved looping with `while` - `take_while` won't fly here since it runs as long as condition is true, whereas i check for the condition to be true to stop looping (i find the location of the cursor, then i know i can stop adding the segments to the path i want to look up in resolver).

PR #1174, opened 2026-01-19
Adjust proc macro server option docs

wawel37:


website/docs/settings.md
```diff
@@ -48,5 +48,8 @@ Absolute path to Cairo core library. It's necessary if the project doesn't use S
 
 Enables on disk cache for procedural macros.
 
+> [!WARNING]
+> With Language Server version 2.15 and above, this option has no effect, as the Proc Macro Cache is no longer experimental and is always enabled.
```
Arcticae: Wait - is this true? Where did this change? (which PR)
Also if it's true, does point below with invalidation still stand?

integraledelebesgue (approved):

PR #1178, opened 2026-01-26
Use new compiler APIs for doc link parsing

Arcticae:
Related Cairo PR: https://github.com/starkware-libs/cairo/pull/9540 (adds the query we use here, consolidates API)

### DIFF
--- a/src/ide/navigation/goto_definition.rs
+++ b/src/ide/navigation/goto_definition.rs
@@ -1,6 +1,7 @@
 use std::ops::Not;
 
 use cairo_lang_defs::db::DefsGroup;
+use cairo_lang_doc::db::DocGroup;
 use cairo_lang_filesystem::ids::{
     FileId, FileKind, FileLongId, SmolStrId, SpanInFile, VirtualFile,
 };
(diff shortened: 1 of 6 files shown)


src/ide/navigation/goto_definition.rs
```diff
@@ -115,46 +94,68 @@ fn try_goto_doc_link_def(
...
 
-    let expr = db.file_expr_syntax(virtual_file).ok()?;
+    let expr = match db.file_expr_syntax(virtual_file) {
+        Ok(expr) => expr,
+        Err(_) => return None,
+    };
     let ast::Expr::Path(expr_path) = expr else {
         return None;
     };
-
```
piotmag769: Why?

src/ide/navigation/goto_definition.rs
```diff
@@ -115,46 +94,68 @@ fn try_goto_doc_link_def(
...
+            "doc-link cursor offsets: cursor_offset={}, dest_span_start={}, dest_span_end={}, relative_cursor_offset={}, dest_text_len={}",
+            cursor_offset.as_u32(),
+            dest_span.start.as_u32(),
+            dest_span.end.as_u32(),
+            relative_cursor_offset,
+            dest_text.len()
+        );
+        let cursor_byte = dest_text.as_bytes().get(relative_cursor_offset as usize).copied();
+        eprintln!("doc-link cursor byte: {:?}, dest_text={:?}", cursor_byte, dest_text);
+    }
```
wawel37: Do we need to keep it?

tests/e2e/goto_definition/doc_links.rs
```diff
@@ -19,19 +19,6 @@ fn crate_path_in_doc_link() {
...
-    test_transform_plain!(GotoDefinition, r"
-    /// See [`  crate::Structu<caret>re`].
-    fn foo() {}
-    struct Structure {}
-    ", @r"
-    /// See [`  crate::Structure`].
-    fn foo() {}
-    struct <sel>Structure</sel> {}
-    ")
-}
```
wawel37: why do we remove this testcase?

Arcticae: We don't support it no more (doesn't make sense to, nobody does it and it would be cumbersome to implement)

Issue #1187, opened 2026-02-12
Add hover for Scarb.toml


FroyaTheHen:
- register Scarb.toml [(here)](https://github.com/software-mansion/cairols/blob/main/src/lsp/capabilities/server.rs)
- add parsing preserving locations in the original file (see [toml_edit](https://crates.io/crates/toml_edit))
- add handle hover logic `fn hover(params: HoverParams, db: &AnalysisDatabase) -> Option<Hover>  ` use [scarb docstring](https://github.com/software-mansion/scarb/issues/2992) as content

PR #1195, opened 2026-02-17
manifest hover

FroyaTheHen:
closes https://github.com/software-mansion/cairols/issues/1187

The idea is to merge the Scarb part first and then update the deps here. 
And I am going to add tests but wanted to expose the changes already 🙏


src/ide/scarb_toml/utils.rs
piotmag769: This module is not needed if you apply the suggestions above

PR #1214, opened 2026-03-06
Add path completions for inner items of traits and impls

### LINKED ISSUE #396: Support all possible elements in autoimports.

wawel37:
resolves https://github.com/software-mansion/cairols/issues/396


integraledelebesgue (approved):

Cargo.lock
```diff
@@ -65,7 +65,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.61.2",
```
Arcticae, line 68: why was this locked to this version?

src/ide/completion/path.rs
```diff
@@ -76,7 +82,86 @@ pub fn path_suffix_completions<'db>(
...
+    else {
+        return Default::default();
+    };
+
+    let is_empty_body_context = is_empty_body_context(db, &ctx.node);
+
+    let (typed_text, last_typed_segment) = match get_typed_text_and_last_segment(db, ctx) {
+        (Some(typed_text), Some(last_typed_segment)) => (typed_text, last_typed_segment),
+        _ if is_empty_body_context => (vec![], SmolStrId::from(db, "")),
+        _ => return Default::default(),
```
Arcticae: This is duplicated a few times in this file

src/ide/completion/path.rs
```diff
@@ -76,7 +82,86 @@ pub fn path_suffix_completions<'db>(
...
+                importable,
+                current_crate,
+                path_str,
+                typed_text.clone(),
+                last_typed_segment,
+            )
+        })
+        .unique_by(|completion| ImportableCompletionItemHashable(completion.clone()))
+        .map(|item| item.item)
+        .collect();
```
Arcticae, line 116: This fragment is duplicated as well

src/ide/completion/path.rs
```diff
@@ -76,7 +82,86 @@ pub fn path_suffix_completions<'db>(
...
+            suffix_completions_by_path(
+                db,
+                ctx,
+                &importables,
+                &typed_text,
+                current_crate,
+                &mut result,
+            );
+        }
+    }
```
Arcticae: This is the only fragment that is special about this function and differentiates it from `importable_path_suffix_completions` - why is it not inside that function (or a separate one)? 

Also this function uses importables so naming is kind of vague here

src/ide/completion/path.rs
```diff
@@ -307,3 +347,306 @@ fn module_item_completion_detail<'db>(
...
+            .all(|(ps, ts)| *ps == ts.to_string(db).as_str());
+        if !matches {
+            continue;
+        }
+
+        let crate_id = trait_id.parent_module(db).owning_crate(db);
+        let is_core = *crate_id.long(db) == CrateLongId::core(db);
+        let relevance = get_item_relevance(true, crate_id == current_crate, is_core);
+
+        // Functions
```
Arcticae: Aren't those repeated?

src/ide/completion/path.rs
```diff
@@ -307,3 +347,306 @@ fn module_item_completion_detail<'db>(
...
+                relevance,
+            }
+        },
+    );
+
+    functions.chain(types).chain(constants).collect()
+}
+
+/// Handles the single-segment suffix case: typing `Name<caret>` proposes
+/// `TraitName::item_name` or `ImplName::item_name` completions for items
```
Arcticae: I wonder if we want to do this, it's a bit awkward both implementation-wise and UX-wise - does RA/any other rust IDE do this? 

I would instead propose items once we have a matching type so the flow would be:

1. I type Name...
2. IDE proposes ImplName/TraitName
3. I choose ImplName
4. I type `::` after `ImplName` - or trigger a completion manually
5. IDE proposes members 

This would be less convoluted i think

Cargo.lock
wawel37: No more changes

src/ide/completion/path.rs
wawel37: changed that to this behaviour

src/ide/completion/path.rs
wawel37: moved fragment to separate function

maciektr:
If I do: 
```cairo
trait Foo {
    fn bar() { }

}
fn test() {
    Foo::<caret>
}
```
I do get completion for `bar(...)`
But if I do
```cairo
trait Foo {
    fn bar() { }

}
fn test() {
    Foo::b<caret>
}
```
I don't get any completions. 
Is this expected? 🤔

wawel37:
@maciektr good catch! fixed that!

maciektr (approved):

src/ide/completion/path.rs
```diff
@@ -30,33 +31,30 @@ use crate::ide::completion::helpers::item::{
...
 use crate::lang::visibility::peek_visible_in_with_edition;
 
-/// Treats provided path as suffix, proposing elements that can prefix this path.
-pub fn path_suffix_completions<'db>(
+/// Proposes importable items whose full path ends with the typed text.
+///
+/// "Suffix" refers to the typed text being matched against the **end** of known importable paths:
+/// e.g. typing `ArrayTrait` surfaces `core::array::ArrayTrait` as a candidate to import.
+/// Used by [`macro_call`] for filtering completions by importable kind.
+pub(crate) fn importable_path_suffix_completions<'db>(
```
Arcticae:
```suggestion
pub fn importable_path_suffix_completions<'db>(
```

Arcticae (approved):

PR #1238, opened 2026-03-24
Add debug lens

piotmag769:
Closes https://github.com/software-mansion/cairo-debugger/issues/17
---
**Stack**:
- #1256
- #1238 ⬅
---

### DIFF
--- a/src/ide/code_lens/debugger.rs
+++ b/src/ide/code_lens/debugger.rs
@@ -1,0 +1,140 @@
+use std::path::PathBuf;
+use std::process::Stdio;
+
+use cairo_lang_defs::db::DefsGroup;
+use cairo_lang_filesystem::db::FilesGroup;
+use lsp_types::notification::ShowMessage;
+use lsp_types::{CodeLens, Command, MessageType, Range, ShowMessageParams, Url};
+
+use crate::ide::code_lens::tests::{
+    TestCodeLensInternal, get_full_path_and_module_id, sanitize_test_case_name,
+};
+use crate::ide::code_lens::{CodeLensInterface, CodeLensInternal, LSCodeLens, make_lens_args};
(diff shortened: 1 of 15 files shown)


src/ide/code_lens/debugger.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::state::State;
+
+#[derive(PartialEq, Clone, Debug)]
+pub struct DebuggerCodeLens {
+    lens: CodeLens,
+}
+
+impl CodeLensInterface for DebuggerCodeLens {
+    fn execute(&self, _file_url: Url, _state: &State, _notifier: &Notifier) -> Option<()> {
+        todo!()
```
piotmag769: Scaffold scaffold - this will do sth like create `.vscode/launch.json` and then launch the debug session in vscode. It will only do so if the client registers custom capabilities for this (so it will only work for vscode). In other cases a notification about using vscode/how to connect to the debugger with a custom client will be displayed (same pattern as in other code lens)

src/ide/code_lens/debugger.rs
```diff
@@ -0,0 +1,127 @@
...
+    db: &AnalysisDatabase,
+    url: Url,
+    test_code_lenses: &[TestCodeLensInternal],
+) -> Option<Vec<DebuggerLensInternal>> {
+    let file = db.file_for_url(&url)?;
+
+    let main_module = *db.file_modules(file).ok()?.first()?;
+    let crate_id = main_module.owning_crate(db);
+
+    // This will not work with crate renames in `Scarb.toml`, but there is no better way to do this now.
```
maciektr: We can mark it with `TODO(software-mansion/scarb#2347)` for future reference I guess

integraledelebesgue (approved):

PR #1262, opened 2026-04-10
Fix dangling proc macro client

### LINKED ISSUE #1057: Fix the dangling proc macro client
`ProcMacroClient` lives in more than one place at the time in the error handler, violating the invariant.

That causes `ProcMacroClientController` to panic on receiving errors from PMS:
https://docs.swmansion.com/maat/release-dev-2025-10-23-0.51.1/pretty_assertions-0.1.0/logs.txt

wawel37:
Tested it with PMS throwing error on every request. With this change we are not getting any "only one strong reference to client is expected at this point" errors, so the client is dropped as it should be, even after editor changes it still works fine

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -399,10 +399,16 @@ impl Backend {
         requester: &mut Requester<'_>,
         _: Responder,
     ) {
-        if let ServerStatus::Connected(client) =
-            state.db.proc_macro_input().proc_macro_server_status(&state.db)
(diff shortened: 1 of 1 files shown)


integraledelebesgue (approved):

maciektr (approved):

piotmag769 (commented):
Sneaky one, good catch!

PR #1264, opened 2026-04-10
Migrate to granular file inputs

integraledelebesgue:


integraledelebesgue:

- #1269
    - #1263
        - #1264 ◀
            - #1265
                - #1266

src/state.rs
```diff
@@ -71,17 +79,144 @@ impl State {
...
+        for (uri, text) in open_file_texts.iter().chain(
+            pending_open_file_overrides
+                .iter()
+                .filter(|(uri, _)| !open_file_texts.contains_key(*uri)),
+        ) {
+            let Some(file_input) =
+                db.file_for_url(uri).map(|file_id| db.file_input(file_id).clone())
+            else {
+                continue;
+            };
```
piotmag769, line 212:
```suggestion
            let Some(file_input) = db.file_for_url(uri)
            else {
                continue;
            };
```
Just call `file_input` later

src/state.rs
```diff
@@ -71,17 +79,144 @@ impl State {
...
+
+            let current_content = db.file_for_url(uri).and_then(|file_id| db.file_content(file_id));
```
piotmag769, line 214:
```suggestion
            let current_content = db.file_content(file_id);
```

src/state.rs
```diff
@@ -71,17 +79,144 @@ impl State {
...
+
+    fn reconcile_open_file_overrides_into_db(
+        db: &mut AnalysisDatabase,
+        open_file_texts: &HashMap<Url, Arc<str>>,
+        pending_open_file_overrides: &HashMap<Url, Arc<str>>,
+    ) {
+        for (uri, text) in open_file_texts.iter().chain(
+            pending_open_file_overrides
+                .iter()
+                .filter(|(uri, _)| !open_file_texts.contains_key(*uri)),
```
piotmag769, line 206: Why filter these? Seems like a logical error - you skip some overrides this way

src/state.rs
```diff
@@ -71,17 +79,144 @@ impl State {
...
+            .collect::<Vec<_>>();
+        for (uri, text) in pending {
+            if self.resolve_file_input(&uri).is_some() {
+                self.apply_open_file_text(uri, text);
+            }
+        }
+    }
+
+    pub fn reconcile_open_file_overrides_in_db(&mut self) {
+        Self::reconcile_open_file_overrides_into_db(
```
piotmag769, line 159: Why not just inline this static method?

src/state.rs
```diff
@@ -71,17 +79,144 @@ impl State {
...
+            &mut self.db,
+            &self.open_file_texts,
+            &self.pending_open_file_overrides,
+        );
+    }
+
+    fn resolve_file_input(&self, uri: &Url) -> Option<cairo_lang_filesystem::ids::FileInput> {
```
piotmag769, line 166:
```suggestion
    fn file_input(&self, uri: &Url) -> Option<cairo_lang_filesystem::ids::FileInput> {
```
[nit]

src/state.rs
```diff
@@ -60,6 +66,8 @@ impl State {
         Self {
             db: AnalysisDatabase::new(),
             open_files: Default::default(),
+            open_file_texts: Default::default(),
+            pending_open_file_overrides: Default::default(),
```
piotmag769, line 70: I don't understand the purpose of this field. You only put stuff into it when `resolve_file_input` fails. This is the case ONLY if `db.file_for_url` fails. But then you need to call `db.file_for_url` again for the same url to even apply the pending overrides. I don't see a case where `db.file_for_url` will return `None` and then `Some` for the same arg - or am I wrong?

src/state.rs
```diff
@@ -71,17 +79,144 @@ impl State {
...
+            .map(|(uri, text)| (uri.clone(), text.clone()))
+            .collect::<Vec<_>>();
+        for (uri, text) in pending {
+            if self.resolve_file_input(&uri).is_some() {
+                self.apply_open_file_text(uri, text);
+            }
+        }
+    }
+
+    pub fn reconcile_open_file_overrides_in_db(&mut self) {
```
piotmag769, line 158: I don't understand the purpose of this method. You apply changes to db on each `didOpen`/`didChange`/`didClose` anyways - why is it needed?

src/project/mod.rs
```diff
@@ -116,82 +116,85 @@ impl ProjectController {
...
+                    } else {
+                        state.analysis_progress_controller.project_model_loaded();
+                    }
+                }
+            };
         }
 
-        // Drop mut ref so we can obtain snapshot.
-        let _ = db;
+        state.apply_pending_open_file_overrides();
```
piotmag769, line 197: Why here?

src/project/model/mod.rs
```diff
@@ -106,6 +106,8 @@ impl ProjectModel {
         db: &mut AnalysisDatabase,
         proc_macro_controller: &ProcMacroClientController,
     ) {
+        db.clear_generated_file_contents();
```
piotmag769, line 109: Why?

src/lang/db/mod.rs
```diff
@@ -72,6 +84,102 @@ impl AnalysisDatabase {
...
+                Some((file_input, (**content).clone()))
+            })
+            .collect()
+    }
+
+    pub fn restore_open_file_overrides(&mut self, overrides: OrderedHashMap<FileInput, Arc<str>>) {
+        for (file_input, content) in overrides {
+            self.set_on_disk_file_content_for_input(file_input, Some(content));
+        }
+    }
```
piotmag769, line 181: Just inline imo

src/server/schedule/task.rs
```diff
@@ -18,7 +18,7 @@ type LocalFn<'s> = Box<dyn FnOnce(&State, MetaState, Notifier, &mut Requester<'_
 
 type BackgroundFn = Box<dyn FnOnce(Notifier, Responder) + Send + 'static>;
 
-pub type BackgroundFnBuilder<'s> = Box<dyn FnOnce(&State, MetaState) -> BackgroundFn + 's>;
+pub type BackgroundFnBuilder<'s> = Box<dyn FnOnce(&mut State, MetaState) -> BackgroundFn + 's>;
```
piotmag769, line 21: Why does this need access to mutable state?

src/state.rs
```diff
@@ -71,17 +79,144 @@ impl State {
...
+                        || !text.is_char_boundary(end)
+                    {
+                        error!(%uri, start, end, text_len = text.len(), "invalid change range");
+                        return;
+                    }
+                    text.replace_range(start..end, &change.text);
+                } else {
+                    text = change.text;
+                }
+            }
```
piotmag769, line 127: Out of scope: this is fine for the future, but we don't support incremental text change anyways, check [here](https://github.com/software-mansion/cairols/blob/ef80d4e1233e214e4a6f6225e7003682fa913d23/src/lsp/capabilities/server.rs#L88)

piotmag769 (changes requested):

src/state.rs
piotmag769: Now I know - but check https://github.com/software-mansion/cairols/pull/1264#discussion_r3073800648

src/lang/db/mod.rs
```diff
@@ -72,6 +84,102 @@ impl AnalysisDatabase {
...
+
+    fn bump_file_contents_revision(&mut self) {
+        let next_revision = files_group_input(self).file_contents_revision(self).saturating_add(1);
+        files_group_input(self)
+            .set_file_contents_revision(self)
+            .with_durability(Durability::HIGH)
+            .to(next_revision);
+    }
+    fn ensure_file_contents_handle_for_input(&mut self, file_input: FileInput) -> FileContents {
```
piotmag769, line 103: [nit] All private should be lower than pub

src/ide/format/documents.rs
```diff
@@ -15,29 +21,49 @@ pub fn format_document(
     let db = &state.db;
     let file_uri = params.text_document.uri;
     let file = db.file_for_url(&file_uri)?;
+    let content = state
+        .open_file_texts
+        .get(&file_uri)
+        .map(|text| text.to_string())
+        .or_else(|| db.file_content(file).map(str::to_owned))?;
```
piotmag769, line 28: Why this way? Is it noticeably faster then just `db.fle_content`?

PR #1267, opened 2026-04-10
Reimplement type info

### LINKED ISSUE #108: Bring back "Type Info"  feature
@mkaput commented on <time datetime="2024-05-09T11:13:54Z">9.05.2024, 11:13:54</time>:
> CairoLS used to show expression type on hover.
> This was changed in https://github.com/starkware-libs/cairo/commit/ba74ab11b8c678f97bf3e12d7145f7e01abd1eb1, which actually is a better goal, but it'd be good to bring back the old behaviour with proper UX.
> IntelliJ Reference: https://www.jetbrains.com/help/idea/viewing-reference-information.html#type-info

maciektr:
closes https://github.com/software-mansion/cairols/issues/108

### DIFF
--- a/src/ide/hover/mod.rs
+++ b/src/ide/hover/mod.rs
@@ -1,3 +1,5 @@
+use cairo_lang_filesystem::db::get_originating_location;
+use cairo_lang_filesystem::ids::SpanInFile;
 use cairo_lang_semantic::lsp_helpers::LspHelpers;
 use cairo_lang_syntax::node::ast::{TerminalIdentifier, TerminalUnderscore};
 use cairo_lang_syntax::node::{SyntaxNode, TypedSyntaxNode};
@@ -23,22 +25,39 @@ pub fn hover(params: HoverParams, db: &AnalysisDatabase) -> Option<Hover> {
 
     let resultants = db.get_node_resultants(node)?;
 
-    let hover_content = resultants
+    // Try standard renderers first (identifiers, literals, keywords, underscores).
+    let content = resultants
         .iter()
         .filter_map(|node| render_hover(db, *node))
-        .collect::<OrderedHashSet<_>>() // Deduplicate so we want display doubled hover if we point to same item from few resultants.
+        .collect::<OrderedHashSet<_>>() // Deduplicate so we don't display doubled hover if we point to same item from few resultants.
         .into_iter()
-        .reduce(|value1, value2| format!("{value1}\n{RULE}{value2}"))?;
+        .reduce(|value1, value2| format!("{value1}\n{RULE}{value2}"));
 
-    let hover = Hover {
-        contents: HoverContents::Markup(MarkupContent {
-            kind: MarkupKind::Markdown,
-            value: hover_content,
-        }),
-        range: node.span_without_trivia(db).position_in_file(db, file_id).map(|p| p.to_lsp()),
+    // Fallback: show the type of the expression at the cursor position, highlighting the
+    // full expression rather than just the hovered token.
+    let (content, node) = match (content, node) {
+        (Some(content), node) => (content, node),
+        _ => resultants.iter().find_map(|node| render::type_info(db, *node))?,
     };
 
-    Some(hover)
+    // Map expanded nodes back to their originating source span so hover highlights work in user
+    // code even when type info comes from a virtual file.
+    let SpanInFile { file_id: range_file_id, span } = get_originating_location(
+        db,
+        SpanInFile { file_id: node.stable_ptr(db).file_id(db), span: node.span_without_trivia(db) },
+        None,
+    );
+    let range = (range_file_id == file_id)
+        .then(|| span.position_in_file(db, file_id).map(|p| p.to_lsp()))
+        .flatten();
+
+    Some(Hover {
+        contents: HoverContents::Markup(MarkupContent {
+            kind: MarkupKind::Markdown,
+            value: content,
+        }),
+        range,
+    })
 }
 
 fn render_hover<'db>(db: &'db AnalysisDatabase, node: SyntaxNode<'db>) -> Option<String> {
--- a/src/ide/hover/render/expression.rs
+++ b/src/ide/hover/render/expression.rs
@@ -1,0 +1,83 @@
+use cairo_lang_defs::ids::{LookupItemId, ModuleItemId};
+use cairo_lang_doc::db::DocGroup;
+use cairo_lang_doc::documentable_item::DocumentableItemId;
+use cairo_lang_semantic::items::function_with_body::{
+    FunctionWithBodySemantic, SemanticExprLookup,
+};
+use cairo_lang_semantic::lookup_item::LookupItemEx;
+use cairo_lang_semantic::lsp_helpers::LspHelpers;
+use cairo_lang_semantic::{ConcreteTypeId, TypeId, TypeLongId};
+use cairo_lang_syntax::node::ast::{Expr, ExprInlineMacro};
+use cairo_lang_syntax::node::{SyntaxNode, TypedSyntaxNode};
+use indoc::formatdoc;
+use itertools::Itertools;
+
+use crate::ide::format::types::format_type;
+use crate::lang::db::{AnalysisDatabase, LsSemanticGroup};
+
+/// Finds the innermost expression containing the given node and renders its inferred type as a
+/// hover popup. Returns the hover content string and the expression's syntax node (used for
+/// determining the highlight range).
+#[tracing::instrument(level = "trace", skip_all)]
+pub fn type_info<'db>(
+    db: &'db AnalysisDatabase,
+    node: SyntaxNode<'db>,
+) -> Option<(String, SyntaxNode<'db>)> {
+    let importables = db.visible_importables_from_module(db.find_module_containing_node(node)?)?;
+
+    let (expr_node, type_id) = node
+        .ancestors_with_self(db)
+        // Walk up to and including any enclosing inline macro. For nodes already in an expansion
+        // file there is no ExprInlineMacro ancestor; for nodes in the source file inside a macro
+        // argument token tree this prevents surfacing types of unrelated outer expressions.
+        .take_while_inclusive(|n| ExprInlineMacro::cast(db, *n).is_none())
+        .filter_map(|n| Expr::cast(db, n))
+        .find_map(|expr| {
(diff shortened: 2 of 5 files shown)


src/ide/hover/mod.rs
```diff
@@ -23,22 +23,35 @@ pub fn hover(params: HoverParams, db: &AnalysisDatabase) -> Option<Hover> {
...
+            }),
+            range: node.span_without_trivia(db).position_in_file(db, file_id).map(|p| p.to_lsp()),
+        });
+    }
 
-    let hover = Hover {
+    // Fallback: show the type of the expression at the cursor position, highlighting the
+    // full expression rather than just the hovered token.
+    let (type_content, expr_node) =
+        resultants.iter().find_map(|node| render::type_info(db, *node))?;
```
maciektr, line 46: @piotmag769 do you think that find_map is rational here? 🤔

piotmag769: I believe it is - a syntactic expression mapping to multiple semantic expressions of different types is a rare occurance

piotmag769 (commented):
I guess UX-wise that's the best we can do without extending LSP protocol (hover params contain only cursor, not a selection)

src/ide/hover/render/expression.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    Some((content, expr_node))
+}
+
+fn find_expr_type<'db>(db: &'db AnalysisDatabase, node: SyntaxNode<'db>) -> Option<TypeId<'db>> {
+    let function_id = db.find_lookup_item(node)?.function_with_body()?;
+    let expr = Expr::from_syntax_node(db, node);
+    let expr_id = db.lookup_expr_by_ptr(function_id, expr.stable_ptr(db)).ok()?;
+    Some(db.expr_semantic(function_id, expr_id).ty())
+}
```
wawel37, line 83: If this returns none if there is no body, maybe we should cover that with test case that doesnt display hover on function-outside expression?

wawel37 (commented):
Please cover cases for inline macro expressions

src/ide/hover/render/expression.rs
maciektr: Done!

maciektr:
@wawel37 please see the changes made. 
Please note, that built-in inline macros won't work for now - fixing this will require changes in the Cairo compiler regarding what code mappings do we emit from them.

maciektr:

- #1267 ◀
    - #1280

wawel37 (approved):

src/ide/hover/render/expression.rs
```diff
@@ -0,0 +1,71 @@
...
+use itertools::Itertools;
+
+use crate::ide::format::types::format_type;
+use crate::lang::db::{AnalysisDatabase, LsSemanticGroup};
+
+/// Finds the innermost expression containing the given node and renders its inferred type as a
+/// hover popup. Returns the hover content string and the expression's syntax node (used for
+/// determining the highlight range).
+///
+/// This is used as a fallback when no other hover renderer handles the hovered position.
```
Arcticae:
```suggestion
```

Arcticae: I don't know if we really need to state that

tests/e2e/hover/type_info.rs
```diff
@@ -0,0 +1,328 @@
...
+    """
+    popover = """
+    ```cairo
+    u32
+    ```
+    """
+    "#)
+}
+
+/// Hover on closing paren of a function call shows the return type.
```
Arcticae, line 49: This one is a bit weird, why wouldn't it show the function type as a whole?

tests/e2e/hover/type_info.rs
```diff
@@ -0,0 +1,328 @@
...
+    """
+    popover = """
+    ```cairo
+    (u32, u64)
+    ```
+    """
+    "#)
+}
+
+/// Hover on struct constructor brace shows the type of the constructed struct.
```
Arcticae: This one should also show the whole type with the `x: u32` etc.

tests/e2e/hover/type_info.rs
```diff
@@ -0,0 +1,328 @@
...
+    """
+    popover = """
+    ```cairo
+    u32
+    ```
+    """
+    "#)
+}
+
+/// Hover on an operator inside a built-in inline macro shows no type info — built-in macros use
```
Arcticae, line 340: Can we leave a TODO to fix that in the compiler first (is this something they would even accept?)

src/ide/hover/mod.rs
```diff
@@ -23,22 +23,43 @@ pub fn hover(params: HoverParams, db: &AnalysisDatabase) -> Option<Hover> {
...
+
+    // The expr_node may be from an expansion (virtual) file; only compute a range when it is
+    // in the same file as the hover request, so we don't report a bogus source position.
+    let range = if expr_node.stable_ptr(db).file_id(db) == file_id {
+        expr_node.span_without_trivia(db).position_in_file(db, file_id).map(|p| p.to_lsp())
+    } else {
+        None
+    };
```
Arcticae: [nit]: We could de-duplicate this logic and only insert the `MarkupContent`'s value, i think we want this condition to hold at all times

tests/e2e/hover/type_info.rs
piotmag769: This was fixed in https://github.com/software-mansion/cairo-language-common/pull/16 I believe

tests/e2e/hover/type_info.rs
piotmag769: Imo its correct - this is not a hover on function name, but on expression that is a function call

tests/e2e/hover/type_info.rs
```diff
@@ -0,0 +1,354 @@
...
+    popover = """
+    ```cairo
+    u64
+    ```
+    """
+    "#)
+}
+
+#[test]
+fn type_info_inside_proc_macro_inline_macro() {
```
piotmag769, line 298: Why no highlight there?

maciektr: Fixed with the change we discussed earlier, will push soon, thx!

tests/e2e/hover/type_info.rs
maciektr: I am kind of split between the two. We could treat () in function call as expression, or as part of function declaration 🤔

src/ide/hover/render/expression.rs
```diff
@@ -0,0 +1,91 @@
...
+pub fn type_info<'db>(
+    db: &'db AnalysisDatabase,
+    node: SyntaxNode<'db>,
+) -> Option<(String, SyntaxNode<'db>)> {
+    let importables = db.visible_importables_from_module(db.find_module_containing_node(node)?)?;
+
+    let (expr_node, type_id) = node
+        .ancestors_with_self(db)
+        // Walk up to and including any enclosing inline macro. For nodes already in an expansion
+        // file there is no ExprInlineMacro ancestor; for nodes in the source file inside a macro
```
piotmag769, line 31: >For nodes already in an expansion

Now that I think about it - we are in the expansion there, no (since we operate on resultants)? There can be no non-top level inline macros left after expansion since we would fall into infinite loop. 

So the question is: does this bring anything? 🤔

src/ide/hover/render/expression.rs
```diff
@@ -0,0 +1,91 @@
...
+        String,
+    >,
+) -> String {
+    if let TypeLongId::Concrete(ConcreteTypeId::Struct(concrete_struct_id)) = ty.long(db)
+        && let Some(signature) = db.get_item_signature(DocumentableItemId::LookupItem(
+            LookupItemId::ModuleItem(ModuleItemId::Struct(concrete_struct_id.struct_id(db))),
+        ))
+    {
+        return signature;
+    }
```
piotmag769, line 70: Shouldn't we try to resolve the concrete type id to its definition (we already have it tbh) and then fall back to our logic from [here](https://github.com/software-mansion/cairols/blob/670e3740200eaca1524dbd51b4914ba8cafaeaab/src/ide/hover/render/definition.rs#L33)?

Maybe it's excessive though and special case for struct is enough

src/ide/hover/render/expression.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    format_type(db, ty, importables, None)
+}
+
+fn find_expr_type<'db>(db: &'db AnalysisDatabase, node: SyntaxNode<'db>) -> Option<TypeId<'db>> {
+    // find_lookup_item works for source-file nodes. Expansion-file nodes (e.g. the result of
+    // proc-macro expansion) have no function items in their AST, so we navigate to the parent
+    // file via code mappings to get the enclosing function_id there.
```
piotmag769: >Expansion-file nodes (e.g. the result of proc-macro expansion) have no function items in their AST

How is that true? 

Btw just use `db.collect_lookup_items_with_parent_files`

tests/e2e/hover/type_info.rs
Arcticae: This is weird from user perspective tbh, since it punishes you for not hovering correctly.

Arcticae: Let's create a follow-up, resolving

Arcticae (approved):

src/ide/hover/render/expression.rs
maciektr: Fixed!

src/ide/hover/render/expression.rs
maciektr: Hmm, do we want it to be as informative as the definition is?
image

piotmag769: It was just an idea, I'm fine with both tbh

maciektr: @Arcticae any opinion? :thinking: 
Personally, I prefer just names, it does not feel especially appropriate to show full definition :thinking:

Arcticae: yeah the definition is a bit excessive, but this doesnt matter that much

src/ide/hover/render/expression.rs
maciektr: Discussed offline - this is not really true.

piotmag769 (approved):

PR #1268, opened 2026-04-10
Find_ref and rename for use aliases.

### LINKED ISSUE #731: `find_ref` and `rename` ignore use aliases (`as`)
Example:
```cairo
use xyz::A as B;

mod xyz {
    pub type A = u8;
}

fn f() {
    let _x: <caret>B = 5;
    let _y: B = 6;
}
```

Both find references and rename don't take `B` from `f` into account

maciektr:
closes #731

- **Write tests**
- **Fix find_references and rename to treat use aliases as own symbols**

### DIFF
--- a/src/lang/defs/finder.rs
+++ b/src/lang/defs/finder.rs
@@ -2,7 +2,7 @@ use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_defs::ids::{
     EnumLongId, GenericTypeId, ImplDefLongId, ImplItemId, LanguageElementId, LookupItemId,
     MacroCallLongId, MemberId, ModuleId, NamedLanguageElementId, StructLongId, SubmoduleLongId,
-    TraitItemId, VarId,
+    TraitItemId, UseId, UseLongId, VarId,
 };
 use cairo_lang_filesystem::ids::SmolStrId;
 use cairo_lang_parser::db::ParserGroup;
@@ -60,6 +60,10 @@ pub enum ResolvedItem<'db> {
     ImplItem(ImplItemId<'db>),
     PluginInlineMacro(SmolStrId<'db>),
     GenericParam(GenericParam<'db>),
+    /// A `use` path leaf with an explicit alias clause (e.g., `use xyz::A as B`).
+    /// The alias name (e.g., `B`) is treated as its own symbol rather than being transparent to
+    /// the underlying item.
+    UseAlias(UseId<'db>),
 }
 
 pub fn find_definition<'db>(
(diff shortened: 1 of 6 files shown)


wawel37 (commented):
how about those cases?
```
      mod xyz {
          pub trait MyTrait<T> {}
      }

      use xyz::MyTrait as AliasedTrait;

      fn foo<T, +<caret>AliasedTrait<T>>() {}
      fn bar<T, +AliasedTrait<T>>() {}
```

maciektr:
@wawel37 good find, fixed!

wawel37 (approved):

src/lang/defs/finder.rs
```diff
@@ -78,10 +83,19 @@ pub fn find_definition<'db>(
...
+        .or_else(|| {
+            let result = try_trait_as_generic_parameter_bound(db, identifier, lookup_items)?;
+            Some(redirect_to_use_alias_if_needed(db, identifier, result))
+        })
         .or_else(|| try_generic_arg(db, identifier, lookup_items))
-        .or_else(|| lookup_resolved_items(db, identifier, lookup_items, resolver_data))
+        .or_else(|| {
+            let result = lookup_resolved_items(db, identifier, lookup_items, resolver_data)?;
+            Some(redirect_to_use_alias_if_needed(db, identifier, result))
+        })
```
Arcticae: [nit]: Wouldn't it be cleaner if it redirected inside of those methods?

Arcticae (approved):

src/lang/defs/finder.rs
maciektr: Done!

tests/e2e/rename/use_alias.rs
piotmag769, line 1: No tests covering macros :/

tests/e2e/find_references/use_alias.rs
piotmag769, line 1: I don' think it should work like this: in this case ALL usages should be found, even the not aliased ones. This is a design matter to discuss though

Arcticae: Interestingly, RA doesn't support much in terms of aliases, not renaming (pic), no finding refs. 
So we're on our own regarding the decisions, and i think the one Maciej made in this one is pretty good, and here's why:
- renaming would be a pain if we found all refs of an alias
- finding the non-aliased references would be MORE confusing IMO, since you're showing something that semantically is correct but is not obvious unless you look up that it's aliased (you would HAVE to show `A` as the declaration if you wanted this behavior you're talking about, so the alias would be shown as a reference only)

image

PR #1273, opened 2026-04-15
Cairo compiler config validation for debugger lens

### LINKED ISSUE #111: Debugger lens - Scarb.toml `[cairo]` config validation in LS
In case the config in invalid - display a message instead of launching the debugger process

piotmag769:
Closes https://github.com/software-mansion/cairo-debugger/issues/111

### DIFF
--- a/src/ide/code_lens/debugger.rs
+++ b/src/ide/code_lens/debugger.rs
@@ -3,6 +3,7 @@ use std::process::Stdio;
 
 use cairo_lang_defs::db::DefsGroup;
 use cairo_lang_filesystem::db::FilesGroup;
+use indoc::indoc;
 use lsp_types::notification::ShowMessage;
 use lsp_types::{CodeLens, Command, MessageType, Range, ShowMessageParams, Url};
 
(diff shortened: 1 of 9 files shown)


src/project/model/configs_registry/package_config.rs
```diff
@@ -53,3 +60,13 @@ fn merge_serde_json_value(a: &mut Value, b: &Value) {
...
+
+fn check_compiler_config_for_debugging(config: &Value) -> bool {
+    let bool_field = |key| config.get(key).and_then(|v| v.as_bool()).unwrap_or(false);
+    let str_field = |key| config.get(key).and_then(|v| v.as_str()).unwrap_or("");
+
+    bool_field("add_functions_debug_info")
+        && bool_field("add_statements_code_locations_debug_info")
+        && bool_field("add_statements_functions_debug_info")
+        && str_field("compiler_optimizations") == "Disabled"
+}
```
piotmag769, line 72: It checks the actual compiler config instead of checking the `Scarb.toml` content, unlike [snforge](https://github.com/foundry-rs/starknet-foundry/issues/4254). This means that if someone has a config like this:
```toml
[cairo] # instead of [profile.dev.cairo]
# all the fields required
```
or like this
```toml
[profile.dev.cairo]
add-statements-locations-debug-info = true # without unstable prefix
# all the fields required
```
the LS validation will pass, but snforge validation will fail. I am fine with that since I want to fix the validation in snforge in a near future and I don't want LS to complain about invalid config when it is fixed in the next snforge versions.

src/lang/diagnostics/refresh.rs
```diff
@@ -72,9 +72,8 @@ fn refresh_file_diagnostics<'db>(
             // not controlled by a user (dependencies from git/package register).
             // Therefore, we filter non-error diagnostics for files residing in Scarb cache
             // and virtual files that are their descendants.
-            let is_dependency = scarb_toolchain.cache_path().is_some_and(|cache_path| {
-                originating_file_path(db, file_id).is_some_and(|p| p.starts_with(cache_path))
-            });
+            let is_dependency = originating_file_path(db, file_id)
```
piotmag769, line 75: Unrelated refactor, noticed it accidentally

src/ide/code_lens/debugger.rs
```diff
@@ -31,7 +32,25 @@ impl CodeLensInterface for DebuggerCodeLens {
         let full_path = sanitize_test_case_name(full_qualified_path.as_ref());
 
         let file_path = file_url.to_file_path().ok()?;
-        let cwd = state.project_controller.configs_registry().manifest_dir_for_file(&file_path)?;
+        let configs = state.project_controller.configs_registry();
+        let cwd = configs.manifest_dir_for_file(&file_path)?;
+
+        if !configs.is_compiler_config_correct_for_debugging_for_file(&file_path)? {
```
wawel37, line 38: Shouldn't it also notify frontend with a message if it returns `None`?

wawel37 (approved):

src/ide/code_lens/debugger.rs
piotmag769: It cannot return `None` if `manifest_dir_for_file` did not return `None` - and it was previously like this as well (and is for test and executable code lens). In general this should never fail, but we can rework it. #1276

PR #1290, opened 2026-05-04
Emit warning diagnostics even on successful metadata runs

### LINKED ISSUE #1189: Add extra keys rejection to `Scarb.toml` parsing
The extra/mismatched (same thing) keys should not be accepted and treated as an error

Arcticae:
Closes #1189 

~~TODO: Pin dev built with https://github.com/software-mansion/scarb/pull/3098 in order for the test to work (otherwise we don't get the warning diagnostic)~~

### DIFF
--- a/src/lang/diagnostics/mod.rs
+++ b/src/lang/diagnostics/mod.rs
@@ -6,7 +6,7 @@ use std::path::Path;
 use cairo_lang_filesystem::ids::FileId;
 use crossbeam::channel::{Receiver, Sender};
 use lsp_types::notification::PublishDiagnostics;
-use lsp_types::{Diagnostic, PublishDiagnosticsParams, Url};
+use lsp_types::{Diagnostic, DiagnosticSeverity, PublishDiagnosticsParams, Url};
 use tracing::{error, trace};
 
 use self::project_diagnostics::ProjectDiagnostics;
@@ -87,15 +87,20 @@ impl DiagnosticsController {
         &mut self,
         root_manifest_path: &Path,
         diagnostics: Vec<ScarbMetadataMessage>,
+        manifest_diagnostic_severity: DiagnosticSeverity,
         db: &AnalysisDatabase,
         notifier: &Notifier,
     ) {
         let Some(root_manifest_url) = Url::from_file_path(root_manifest_path).ok() else {
             return;
         };
-        let diagnostics =
-            scarb_metadata_messages_to_diagnostics(db, diagnostics, root_manifest_path)
-                .unwrap_or_default();
+        let diagnostics = scarb_metadata_messages_to_diagnostics(
+            db,
+            diagnostics,
+            root_manifest_path,
+            manifest_diagnostic_severity,
+        )
+        .unwrap_or_default();
         let diags_to_send = self.update_scarb_manifest_diagnostics(root_manifest_url, diagnostics);
         for (url, diagnostics) in diags_to_send {
             notifier.notify::<PublishDiagnostics>(PublishDiagnosticsParams {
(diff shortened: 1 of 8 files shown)


src/lang/diagnostics/mod.rs
```diff
@@ -87,15 +87,20 @@ impl DiagnosticsController {
         &mut self,
         root_manifest_path: &Path,
         diagnostics: Vec<ScarbMetadataMessage>,
+        manifest_diagnostic_severity: DiagnosticSeverity,
```
maciektr, line 90: Shouldn't we be able to read severity from Scarb, instead of assuming it for all the messages? 🤔

src/toolchain/scarb.rs
```diff
@@ -157,16 +162,29 @@ impl ScarbToolchain {
...
-        MetadataCommand::new()
-            .scarb_path(scarb_path)
-            .manifest_path(manifest_path)
-            .json()
-            .inherit_stderr()
-            .exec()
-            .inspect_err(|err| {
-                error!("{err:?}");
+    ) -> Result<MetadataOutput, MetadataCommandError> {
+        let output = Command::new(scarb_path)
```
maciektr, line 166: Shouldn't we modify scarb_metadata so we still can use MetadataCommand here for deserialization?

src/lang/diagnostics/mod.rs
Arcticae: Scarb side doesn't emit that. It could but i didn't want to change the API for that PR

src/toolchain/scarb.rs
Arcticae: That could be integrated directly there. It would probably require API changes as well if we wanted to emit diagnostics via those methods, so i wasn't sure of the consequences.

src/lang/diagnostics/mod.rs
Arcticae: https://github.com/software-mansion/scarb/issues/3118

src/toolchain/scarb.rs
Arcticae: https://github.com/software-mansion/scarb/issues/3119

src/project/scarb_manifest_diagnostics.rs
```diff
@@ -105,8 +87,42 @@ struct LspScarbDiagnostic {
...
 
-fn metadata_messages_from_ndjson(stdout: &str) -> Vec<ScarbMetadataMessage> {
-    ndjson_values(stdout).filter_map(|value| ScarbMetadataMessage::try_from(&value).ok()).collect()
+pub(crate) fn manifest_diagnostics_from_ndjson(stdout: &str) -> Vec<ScarbMetadataMessage> {
+    ndjson_values(stdout).filter_map(|value| manifest_diagnostic_from_value(&value)).collect()
+}
+
+fn metadata_errors_from_ndjson(stdout: &str) -> Vec<ScarbMetadataMessage> {
+    ndjson_values(stdout).filter_map(|value| metadata_error_from_value(&value)).collect()
+}
```
wawel37, line 96: nit: make the naming consistent, both should be `manifest` or `metadata` imho, as it might be confusing in the future

wawel37 (approved):
Looks good!

src/project/scarb_manifest_diagnostics.rs
Arcticae: The distinction is there for a specific reason - it's called this way because metadata can error for multiple reasons, and manifest diagnostics are a specific thing emitted when checking the metadata. Those are 2 separate concepcts.

maciektr (approved):

PR #1301, opened 2026-05-15
Reload project model when integration test files are created

### LINKED ISSUE #1285: Newly created `tests/*.cairo` file is not recognised until project model is reloaded
1. Make sure there is no `tests/lib.cairo` file in your project.
2. Add `tests/xyz.cairo` file.
3. LSP features don't work inside `tests/xyz.cairo`.
4. Save `Scarb.toml` (`scarb metadata` executes again) - now they do work.

The same thing would happen when e.g. `tests/lib.cairo` is created and it was not there before.

This happens because integration tests act as a separate create inside Cairo compiler db. We get information about the structure of integration tests from `scarb metadata`, or more precisely:
- if they exist at all
- if they do: whether `tests/lib.cairo` exists
- if there is no `tests/lib.cairo`: what are the top level files in `tests/` dir (matching `tests/*.cairo` glob pattern)

If this changes (e.g. a new file `tests/*.cairo` is created when there is no `tests/lib.cairo` OR `tests/lib.cairo` is created and it was not there before OR `tests/lib.cairo` is deleted), the way the entry file to the integration tests crate is handled changes as well, which casuses issues.

We should probably solve it in `DidChangeWatchedFiles` request handler - just reload the project model for the particular crate/workspace (we can do that only for this crate as integration test changes only affect this crate - project model does not have this capability as of now though) if:
- `tests/lib.cairo` is deleted
- `tests/lib.cairo` is created
- `tests/*.cairo` is created and there is no `tests/lib.cairo`

trungkien1992:
Fixes #1285.

Scarb can derive integration-test crates from direct `tests/*.cairo` files, so create/delete file events can change project metadata even when `Scarb.toml` is unchanged. Before this change, CairoLS invalidated the changed Cairo file but did not reload the project model, so newly created integration tests were not analyzed until a broader reload happened.

This keeps the reload intentionally narrow: only direct `tests/*.cairo` create/delete events trigger a project model reload. Regular Cairo file edits still avoid a backend reload, and proc-macro restart remains limited to config-file changes.

Verification:
- `cargo fmt --check`
- `SCARB_BIN="$(asdf where scarb nightly-2026-05-09)/bin" PATH="$SCARB_BIN:$PATH" cargo test scarb_toml_change`

Local note: these tests need the repo-pinned Scarb `nightly-2026-05-09`; an older `scarb 2.8.0` on PATH fails unrelated Scarb/corelib behavior.

### DIFF
--- a/src/server/routing/handlers.rs
+++ b/src/server/routing/handlers.rs
@@ -6,6 +6,7 @@
 // +-----------------------------------------------------+
 
 use std::panic::{AssertUnwindSafe, catch_unwind, resume_unwind};
+use std::path::Path;
 
 use cairo_lang_filesystem::db::FilesGroup;
 use cairo_lang_filesystem::override_file_content;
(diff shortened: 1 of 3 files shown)


piotmag769:
This code is not aligned with the proposed solution in the issue description.

PR #1304, opened 2026-05-19
feat: add non-blocking try_wait on trigger receiver

Arcticae:
Adds `try_wait` / `Inner::try_wait` so the diagnostics controller can
drain queued triggers without blocking — needed for the debounce loop
that collapses rapid refresh requests into one scarb check invocation.


Arcticae:

- #1304 ◀
    - #1305
        - #1306
            - #1309
                - #1307
                    - #1308

PR #1311, opened 2026-05-22
Disposable DB for diagnostics

### LINKED ISSUE #1103: LS is memory hungry
### Problem

It can easily reach 10GB+ of RAM (sometimes we observe ~16GB). It seems each letter typed increases the memory usage, becauce everything is being recalculated.

Let's think about how to remedy this. There were ideas to keep part of the tree on disk (those that typically do not change, so deps of the deps etc), and only keep parts of root + 1-2 subtrees in memory (for faster reads).

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Scarb Version

```text

```

Arcticae:


src/lang/db/migration.rs
Arcticae, line 1: Logic mostly copied over to be reused

src/lang/diagnostics/refresh.rs
```diff
@@ -17,6 +17,7 @@ use crate::toolchain::scarb::ScarbToolchain;
 
 /// Refresh diagnostics and send diffs to the client.
 #[tracing::instrument(skip_all)]
+#[allow(clippy::too_many_arguments)]
```
piotmag769: Why suddenly?

src/lang/db/migration.rs
```diff
@@ -0,0 +1,125 @@
...
+    open_files: &HashSet<Url>,
+    project_controller: &ProjectController,
+    proc_macro_client_controller: &ProcMacroClientController,
+) -> AnalysisDatabase {
+    let mut new_db = AnalysisDatabase::new();
+
+    migrate_default_plugins(&mut new_db, old_db);
+    migrate_proc_macro_state(&mut new_db, old_db);
+    migrate_file_overrides(&mut new_db, old_db, open_files);
+    migrate_crate_model_inputs(&mut new_db, old_db);
```
piotmag769, line 26: The line below does the same, no?

src/lang/diagnostics/mod.rs
```diff
@@ -175,27 +211,36 @@ impl DiagnosticsController {
     }
 }
 
+#[derive(Clone)]
+struct DiagnosticsRunInput {
+    db: AnalysisDatabase,
+    scarb_toolchain: ScarbToolchain,
```
piotmag769: Don't pass it every time, just keep it in the controller thread - it has arc inside it anyways

src/lang/diagnostics/mod.rs
```diff
@@ -43,31 +46,30 @@ pub struct DiagnosticsController {
     //   The trigger MUST be dropped before worker's join handle.
     //   Otherwise, the controller thread will never be requested to stop, and the controller's
     //   JoinHandle will never terminate.
-    trigger: trigger::Sender<StateSnapshot>,
+    trigger: trigger::Sender<DiagnosticsRunInput>,
+    active_diagnostics_db: Arc<Mutex<Option<AnalysisDatabase>>>,
```
piotmag769, line 54: Pls change as discussed offline

wawel37 (changes requested):
Ok, I've tried testing it locally on OZ repo, and I've notice major problem. The Diagnostic DB is being dropped only when the new diagnostic refresh request comes in, which result in us storing the whole db during the idling time. I think it shouldnt be the case, as during the idle we are still taking ~7-8 GB. This implementation only saves us from ever growing salsa db from diagnostics, but during the idle we still consume a lot of memory. Is it possible to somehow drop the diagnostics db after the diagnostic calculation end?

src/lang/db/migration.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    project_controller.migrate_crates_to_new_db(&mut new_db, proc_macro_client_controller);
```
piotmag769: We can remove this call then since we do it in L30

piotmag769 (approved):

Arcticae:
@wawel37 old behavior was restored, you can re-test stuff if you want

src/lang/db/migration.rs
```diff
@@ -0,0 +1,119 @@
...
+        };
+        let file_input = file_id.long(old_db).into_file_input(old_db);
+        if let Some(content) = overrides.get(&file_id) {
+            new_overrides.insert(file_input, content.to_string().into());
+        }
+    }
+    files_group_input(new_db).set_file_overrides(new_db).to(Some(new_overrides));
+}
+
+/// Copies the currently configured project crates without migrating query outputs.
```
piotmag769:
```suggestion
/// Copies the currently configured project crates.
```

src/lang/diagnostics/mod.rs
```diff
@@ -43,7 +46,12 @@ pub struct DiagnosticsController {
     //   The trigger MUST be dropped before worker's join handle.
     //   Otherwise, the controller thread will never be requested to stop, and the controller's
     //   JoinHandle will never terminate.
-    trigger: trigger::Sender<StateSnapshot>,
+    trigger: trigger::Sender<DiagnosticsRunInput>,
+    // This is intentionally shared between the main thread and the diagnostics controller thread:
+    // the main thread must be able to cancel the currently running diagnostics DB as soon as a new
+    // refresh is requested, while the diagnostics thread is the one that sets it at tick start and
+    // clears it after the tick. The mutex protects that single shared slot.
```
piotmag769: I would mention that this is done to reduce memory consumption

src/lang/diagnostics/mod.rs
```diff
@@ -322,4 +381,25 @@ impl DiagnosticsControllerThread {
...
     }
+
+    fn set_active_diagnostics_db(&self, db: AnalysisDatabase) {
+        *self
+            .active_diagnostics_db
+            .lock()
+            .expect("active diagnostics db mutex should never be poisoned") = Some(db);
+    }
+
+    fn clear_active_diagnostics_db(&self, db: AnalysisDatabase) {
```
piotmag769: You can take ref to db to avoid the clone on the caller side

piotmag769 (approved):

src/lang/diagnostics/mod.rs
```diff
@@ -78,9 +89,43 @@ impl DiagnosticsController {
         self.generate_code_complete_receiver.clone()
     }
 
-    /// Schedules diagnostics refreshing on snapshot(s) of the current state.
-    pub fn refresh(&self, state: &State) {
-        self.trigger.activate(state.snapshot());
+    /// Schedules diagnostics refreshing on a fresh diagnostics-only database.
+    #[allow(clippy::too_many_arguments)]
```
wawel37: does it really fire for only 5 args?

src/lang/diagnostics/mod.rs
```diff
@@ -175,14 +220,23 @@ impl DiagnosticsController {
     }
 }
 
+#[derive(Clone)]
```
wawel37: nit: not sure but i dont see it being cloned anywhere

wawel37 (approved):
Looks good! GJ :D

maciektr (approved):
This looks really great! The change in mem usage is real!

PR #1313, opened 2026-05-25
Fix fuzzer with test codelens

wawel37:
https://github.com/software-mansion/cairols/issues/1245


Arcticae (approved):

maciektr (approved):

piotmag769 (changes requested):
This is a regression - there should be no debug lens that launch fuzzer tests (only test cases should have these), now there are

src/ide/code_lens/tests.rs
```diff
@@ -384,20 +383,25 @@ fn is_fuzzer_test(db: &AnalysisDatabase, ptr: SyntaxStablePtrId) -> bool {
         return false;
     };
 
-    // We do not want to skip test cases.
-    // `#[test_case]` generates a new test with `#[snforge_internal_test_executable]`
-    // directly so it is okay to test for it as an ancestor.
+    // If the originating location is at a #[test_case] attribute, this is a concrete test case
+    // generated by the test_case macro — not a fuzzer. The original code used
+    // `ancestor_of_type::<Attribute>` (which excludes self) for this check, but after a Cairo
+    // bump `find_syntax_node_at_offset` started returning the attribute node itself, so we use
```
piotmag769: Which cairo bump? I don't see any changes related to the code that is used by `find_syntax_node_at_offset` implementation

piotmag769 (commented):
This changes change nothing logic-wise tbh.
Now that I look at the issue I created, it may make no sense - maybe we don't need the ability to run all test cases and fuzzer and the same time?

src/ide/code_lens/tests.rs
wawel37: Removed that

wawel37:
Decided on slack its not needed for now

PR #1315, opened 2026-05-26
Fix macro expansion code action

maciektr:
fixes #1246

- **Upgrade snforge_std to latest**
- **Add test case demonstrating the problem**
- **Fix macro expansion code action**

### DIFF
--- a/src/ide/macros/expand/mod.rs
+++ b/src/ide/macros/expand/mod.rs
@@ -83,28 +83,50 @@ fn expand<'db>(
 
     let mut files = files.into_iter().peekable();
 
-    let mut content = expand_inline(db, start_file, inline_files)?;
-
-    let Some(first_file) = files.peek() else {
-        return Some(content);
+    let first_file = if files.peek().is_some_and(|f| f.as_virtual(db).original_item_removed) {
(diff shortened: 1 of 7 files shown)


src/ide/macros/expand/mod.rs
```diff
@@ -83,28 +83,50 @@ fn expand<'db>(
 
     let mut files = files.into_iter().peekable();
 
-    let mut content = expand_inline(db, start_file, inline_files)?;
-
-    let Some(first_file) = files.peek() else {
-        return Some(content);
+    let first_file = if files.peek().is_some_and(|f| f.as_virtual(db).original_item_removed) {
```
piotmag769:
```suggestion
    let maybe_file_replacing_og_item = if files.peek().is_some_and(|f| f.as_virtual(db).original_item_removed) {
```
or sth like that

piotmag769 (approved):
Seems g logicwise

src/ide/macros/expand/mod.rs
maciektr: Done

Arcticae (approved):

PR #1321, opened 2026-06-02
Swap database on inactivity

Arcticae:

PR #1322, opened 2026-06-02
Remove active time database swapper

### LINKED ISSUE #1312: Check if current swapper logic really fires after some edits or memory spike
Some time ago we implemented a logic for a swapper, that runs the swap under certain conditions. Lately we've noticed that it doesn't fire nearly at all.

AC:
- Check if current logic that determines when we swap the DB still works.
- If not, propose and implement another way to do that logic check

Arcticae:


Arcticae:

- #1321
    - #1322 ◀

piotmag769 (approved):

maciektr (approved):

wawel37 (approved):

PR #1325, opened 2026-06-03
Fix hovers for closure params

### LINKED ISSUE #120: Hovers do not work for closure parameters
### Problem

Goto definition used to work before its refactoring, but hovers have never worked.

I debugged that and the reason for this is that closures are not handled in lookup items finding logic and are not present in `LookupItemId`.

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Scarb Version

```text

```

wawel37:


tests/e2e/hover/closures.rs
Arcticae, line 1: This doesn't test any other param type than felt. Can closure parameters contain other param types as well?

wawel37: Added additional tests for various types in closure

Arcticae (approved):
LGTM

maciektr (approved):

maciektr:
Looks good to me!

PR #1329, opened 2026-06-09
Add ls discovery section to docs

wawel37:


website/docs/language-server-binary-discovery.md
```diff
@@ -0,0 +1,36 @@
...
+It will _not_ fall back to other methods if this path is incorrect!
+If `cairo1.scarbPath` is unset, then it searches in your system's `PATH` environment variable.
+
+## Scarb Discovery via ASDF
+
+> [!NOTE]
+> There is no `asdf` for Windows, so this step will not occur there.
+
+> [!WARNING]
+> You should **not** rely on this behavior; use configuration or the `PATH` environment variable
```
maciektr: I believe most people actually rely on this, as this is the default behaviour when you just install Scarb via starkup/asdf and the extension from store.

website/docs/language-server-binary-discovery.md
```diff
@@ -0,0 +1,36 @@
...
+> instead!
+
+Should the search in `PATH` fail, the extension employs a more involved logic to locate the `asdf`
+version manager tool on your machine. This step accounts for instances where
+[Scarb](https://docs.swmansion.com/scarb/) is installed via `asdf` but `asdf` is not available in
+the system `PATH`.
```
maciektr: > installed via `asdf` but `asdf` is not available in the system `PATH`.

How come? 🤔 
If there is no `asdf` in PATH, how would LS call `asdf`? 🤔

maciektr (changes requested):

website/docs/language-server-binary-discovery.md
wawel37: Removed the warning 👍

website/docs/language-server-binary-discovery.md
wawel37: Specified more clearly that we look inside the shims of asdf default installation directory (`~/.asdf/shims/scarb`)

maciektr (approved):

Arcticae (approved):

PR #1330, opened 2026-06-10
Add `replace` type quick fixes for invalid `inherits` in Scarb manifest

Arcticae:
We have only a couple of supported inheritance values anyways so we can propose correct ones on user mistake

### DIFF
--- a/tests/e2e/code_actions/scarb_manifest.rs
+++ b/tests/e2e/code_actions/scarb_manifest.rs
@@ -91,10 +91,12 @@ fn manifest_profile_inheritance_invalid_code_action_endpoint_returns_fixes() {
     let dev_action = find_code_action(code_actions.clone(), "Set `inherits` to `\"dev\"`");
     let dev_text = only_manifest_edit(&dev_action);
     assert!(dev_text.contains(r#"inherits = "dev""#), "{dev_text}");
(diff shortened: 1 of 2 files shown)


maciektr (approved):

tests/e2e/code_actions/scarb_manifest.rs
```diff
@@ -64,23 +64,37 @@ fn manifest_inlining_strategy_conflict_code_action_endpoint_returns_fixes() {
...
+    assert!(!remove_text.contains("inherits"), "{remove_text}");
+    assert!(remove_text.contains("[profile.custom]"), "{remove_text}");
+
+    let dev_action = find_code_action(code_actions.clone(), "Set `inherits` to `\"dev\"`");
+    let dev_text = only_manifest_edit(&dev_action);
+    assert!(dev_text.contains(r#"inherits = "dev""#), "{dev_text}");
+
+    let release_action = find_code_action(code_actions, "Set `inherits` to `\"release\"`");
+    let release_text = only_manifest_edit(&release_action);
+    assert!(release_text.contains(r#"inherits = "release""#), "{release_text}");
```
wawel37, line 98: nit: for those we should also check if the old `inherits = "invalid"` is gone

wawel37 (approved):

PR #1331, opened 2026-06-10
Emit more specific popup for metadata fail with cairo version mismatch

maciektr:
The cairo-version field in manifest may require different scarb/cairo toolchain version than currently available. In this case, emit a more specific popup instead of generic "metadata failed" error

### DIFF
--- a/src/project/mod.rs
+++ b/src/project/mod.rs
@@ -21,7 +21,7 @@ pub use self::scarb_manifest_diagnostics::{
 };
 use self::scarb_manifest_diagnostics::{
     collect_scarb_manifest_diagnostics, manifest_diagnostics_from_ndjson,
-    scarb_metadata_messages_contain_only_errors,
+    scarb_metadata_failed_message,
 };
 use crate::ide::code_lens::FileChange;
 use crate::lang::db::AnalysisDatabase;
@@ -322,9 +322,13 @@ impl ProjectControllerThread {
                     .unwrap_or_else(|error| {
                         let metadata_messages = collect_scarb_manifest_diagnostics(error);
 
-                        if scarb_metadata_messages_contain_only_errors(&metadata_messages) {
-                            self.scarb_toolchain.notify_metadata_failed();
-                        }
+                        // A failed `scarb metadata` means the project could not be loaded, so we
+                        // always inform the user with a popup - the message is tailored to the
+                        // cause (e.g. a Cairo version mismatch) when we can recognize it.
+                        // Located manifest diagnostics are additionally surfaced inline.
+                        self.scarb_toolchain.notify_metadata_failed(scarb_metadata_failed_message(
+                            &metadata_messages,
+                        ));
 
                         ProjectUpdate::ScarbMetadataFailed {
                             manifest_path,
--- a/src/project/scarb_manifest_diagnostics.rs
+++ b/src/project/scarb_manifest_diagnostics.rs
@@ -9,6 +9,9 @@ use serde_json::Value;
 
 use crate::lang::db::AnalysisDatabase;
 use crate::lang::lsp::Utf8Span;
+use crate::toolchain::scarb::{
+    SCARB_METADATA_CAIRO_VERSION_MISMATCH_MESSAGE, SCARB_METADATA_FAILED_MESSAGE,
+};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ScarbMetadataMessage {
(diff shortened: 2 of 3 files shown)


src/toolchain/scarb.rs
```diff
@@ -187,10 +190,10 @@ impl ScarbToolchain {
         })
     }
 
-    pub fn notify_metadata_failed(&self) {
+    pub fn notify_metadata_failed(&self, message: &str) {
```
Arcticae, line 193: the function no longer signals only failed metadata

src/toolchain/scarb.rs
```diff
@@ -17,6 +17,9 @@ use crate::server::client::Notifier;
 pub const SCARB_TOML: &str = "Scarb.toml";
 pub const SCARB_METADATA_FAILED_MESSAGE: &str =
     "`scarb metadata` failed. Check if your project builds correctly via `scarb build`.";
+pub const SCARB_METADATA_CAIRO_VERSION_MISMATCH_MESSAGE: &str = "`scarb metadata` failed due to a Cairo version requirement mismatch. The Cairo version \
```
Arcticae, line 20: it's not used in this file, it shouldn't be here probably

src/project/scarb_manifest_diagnostics.rs
```diff
@@ -33,9 +36,23 @@ fn diagnostics_to_display(all_messages: Vec<ScarbMetadataMessage>) -> Vec<ScarbM
...
+pub fn scarb_metadata_failed_message(messages: &[ScarbMetadataMessage]) -> &'static str {
+    if messages.iter().any(scarb_metadata_message_is_cairo_version_mismatch) {
+        SCARB_METADATA_CAIRO_VERSION_MISMATCH_MESSAGE
+    } else {
+        SCARB_METADATA_FAILED_MESSAGE
+    }
+}
+
+fn scarb_metadata_message_is_cairo_version_mismatch(message: &ScarbMetadataMessage) -> bool {
+    match message {
```
Arcticae, line 52: [nit]: i'd rather have if let chain here

src/project/scarb_manifest_diagnostics.rs
```diff
@@ -33,9 +36,23 @@ fn diagnostics_to_display(all_messages: Vec<ScarbMetadataMessage>) -> Vec<ScarbM
...
 }
 
-pub fn scarb_metadata_messages_contain_only_errors(messages: &[ScarbMetadataMessage]) -> bool {
-    !messages.is_empty()
-        && messages.iter().all(|message| matches!(message, ScarbMetadataMessage::MetadataError(_)))
+/// Picks the most specific user-facing message for a failed `scarb metadata` invocation.
+///
+/// When the failure is caused by a Cairo version requirement mismatch, returns a dedicated
+/// message pointing the user at the root cause instead of the generic one.
+pub fn scarb_metadata_failed_message(messages: &[ScarbMetadataMessage]) -> &'static str {
```
Arcticae, line 43: `ScarbMetadataMessage` prescence doesn't mean that a metadata failed - it can contain diagnostics only which are non fatal. this function can easily lead to misunderstanding of the metadata internals

src/project/mod.rs
```diff
@@ -322,9 +322,13 @@ impl ProjectControllerThread {
...
                         let metadata_messages = collect_scarb_manifest_diagnostics(error);
 
-                        if scarb_metadata_messages_contain_only_errors(&metadata_messages) {
-                            self.scarb_toolchain.notify_metadata_failed();
-                        }
+                        // A failed `scarb metadata` means the project could not be loaded, so we
+                        // always inform the user with a popup - the message is tailored to the
+                        // cause (e.g. a Cairo version mismatch) when we can recognize it.
+                        // Located manifest diagnostics are additionally surfaced inline.
+                        self.scarb_toolchain.notify_metadata_failed(scarb_metadata_failed_message(
```
Arcticae, line 329: This can probably be done easily inside metadata, exposed in invocation result somehow and not leak metadata handling logic outside to project layer.

Arcticae (changes requested):

src/toolchain/scarb.rs
maciektr: I don't think I understand what you are trying to say 🤔 Invalid cairo is still a metadata fail.

src/project/scarb_manifest_diagnostics.rs
```diff
@@ -33,9 +36,23 @@ fn diagnostics_to_display(all_messages: Vec<ScarbMetadataMessage>) -> Vec<ScarbM
...
+    if messages.iter().any(scarb_metadata_message_is_cairo_version_mismatch) {
+        SCARB_METADATA_CAIRO_VERSION_MISMATCH_MESSAGE
+    } else {
+        SCARB_METADATA_FAILED_MESSAGE
```
wawel37, line 47: Correct me if im wrong, but not all metadata messages are strictly "fails".

src/toolchain/scarb.rs
wawel37, line 20: nit: i'd move those 2 to `scarb_manifest_diagnostics`

src/toolchain/scarb.rs
Arcticae: I mean read what the code does right now - it doesn't send metadata failure message. It sends any message you pass it

PR #1349, opened 2026-06-30
Bump Cairo

integraledelebesgue:


src/lang/defs/finder.rs
```diff
@@ -449,7 +449,9 @@ fn try_member<'db>(
         current_node = current_node.parent(db).unwrap();
     }
 
-    let member_id = expr_member_access.member;
+    let MemberAccessKind::Struct { member_id, .. } = expr_member_access.kind else {
+        return None;
+    };
```
piotmag769, line 454: An idea (kinda related to this place in code) - do we want to support hovers on tuple access? E.g. when hovering over 
```
let x = (1_u8, 2_u32);
let y = x.0<caret> + 5;
```
we could show the type

```
u8
```

Similarly we could support find ref and goto.

WDYT?

piotmag769 (approved):

wawel37 (approved):

PR #1353, opened 2026-07-01
Add tests for code lenses for proc-macro-generated tests

integraledelebesgue:

PR #1338, opened 2026-06-17
Add tests for user-defined inline macros

integraledelebesgue:

PR #1337, opened 2026-06-17
Fix expansion for user-defined inline macros

integraledelebesgue:

PR #1354, opened 2026-07-01
Fix code lenses for proc-macro-generated modules

### LINKED ISSUE #683: Code lenses in proc macro controlled code

integraledelebesgue:
Closes: #683


integraledelebesgue:

- #1337
    - #1338
        - #1353
            - #1354 ◀
                - #1362

src/ide/code_lens/tests.rs
```diff
@@ -89,9 +99,25 @@ pub fn get_full_path_and_module_id<'db>(
...
+
+    if let Some(resolved) = resolved {
+        return Some(resolved);
+    }
+
+    // Fallback for items inside a macro-generated `mod`: `get_node_resultants` does not descend
+    // into the generated module's contents, so the click cannot be re-resolved to the `mod` (or
+    // the tests inside it). Use the path computed at discovery time and the file's owning crate
+    // (all the runner detection below needs).
+    let module_id = *db.file_modules(file).ok()?.first()?;
```
wawel37, line 96: what if the macro generates multiple modules? or it's not possible?

integraledelebesgue: It works properly. I added a test for that

tests/e2e/code_lens/proc_macro.rs
```diff
@@ -87,14 +87,13 @@ fn handwritten_module_aggregates_generated_test() {
...
-// the click cannot be resolved back to a macro-generated `mod` item.
+// A whole module of tests generated by an inline macro: the module gets a `Run tests` lens and
+// the inner test gets Run/Debug lenses, all on the macro call line. Executing the generated
+// module's lens resolves via the stored-path fallback.
 #[test]
 fn inline_macro_generates_test_module() {
     test_transform!(test_code_lens_snforge_with_macros, r#"
-    test_module_generating_inline_macro_v2!();
-
-    fn not_a_test() {}<caret>
```
Arcticae, line 97: [nit]: why delete that?

src/ide/code_lens/tests.rs
```diff
@@ -72,15 +78,19 @@ pub fn get_full_path_and_module_id<'db>(
...
 
     let file = db.file_for_url(file_url)?;
-    let span = TextPositionSpan::offset_in_file(lens.range.to_cairo(), db, file)?;
-    // When creating range we have to use [`SyntaxNode::span_without_trivia`] to place lens next to attribute
-    let node = db.widest_node_within_span_without_trivia(file, span)?;
 
-    let (full_qualified_path, module_id) =
+    // Re-resolve the clicked node to recover the (possibly updated) path and its module. This
+    // works for hand-written tests and for proc-macro-generated functions (expanded in place).
+    let resolved = (|| {
```
Arcticae: a regular block yielding a value would be fine here - no need for a lambda

Arcticae: also you could combine this into `if let resolved = {...} else {...}` but that's a matter of preference here

src/ide/code_lens/tests.rs
```diff
@@ -89,9 +99,25 @@ pub fn get_full_path_and_module_id<'db>(
...
-        })?;
+        })
+    })();
+
+    if let Some(resolved) = resolved {
+        return Some(resolved);
+    }
+
+    // Fallback for items inside a macro-generated `mod`: `get_node_resultants` does not descend
+    // into the generated module's contents, so the click cannot be re-resolved to the `mod` (or
```
Arcticae: >`get_node_resultants` does not descend into the generated module's contents

that's weird behavior to implement into code lens - shouldn't this be fixed in resultants directly?

src/ide/code_lens/tests.rs
Arcticae: what in case of nested generated modules? (i.e. 2 levels of nesting the generation)?

i know it's bizarre but this might happen

src/ide/code_lens/tests.rs
integraledelebesgue: This comment was wrong, I corrected it. Now nothing really needs upstreaming

src/ide/code_lens/tests.rs
integraledelebesgue: Good catch. It works, I added a test: https://github.com/software-mansion/cairols/blob/fd96154fec017a5890e8773134d61bee9226b37f/tests/e2e/code_lens/proc_macro.rs#L371-L420

tests/e2e/code_lens/proc_macro.rs
integraledelebesgue: It was an unnecessary placeholder

src/ide/code_lens/tests.rs
integraledelebesgue: Not really - I have several `?` inside the lambda. This is more readable than a chain of `if let Some(...)`. A dedicated function for this would be excessive

Arcticae: I would rather have a dedicated function: 
You execute one task, leverage function-like behavior, document it over the call
Only argument against is lack of reuse which is not enough for me

Arcticae (approved):

src/ide/code_lens/tests.rs
integraledelebesgue: Agree, done

wawel37 (approved):

PR #1363, opened 2026-07-06
Fix impl methods completions

### LINKED ISSUE #1314: Impl methods aren't proposed in completions
### Problem

It seems that the completion should fire here with the method name but it doesn't
Image

The implementation looks like this basically (it's from alexandria)
```
pub impl MergeSort of Sortable {
    // Merge Sort
    /// #### Arguments
    /// * `arr` - Array to sort
    /// #### Returns
    /// * `Array<T>` - Sorted array
    fn sort<T, +Copy<T>, +Drop<T>, +PartialOrd<T>>(mut array: Span<T>) -> Array<T> {
        let len = array.len();
        // ...
        merge_recursive(sorted_left, sorted_right, ref result_arr, 0, 0);
        result_arr
    }
}
```

### Steps

_No response_

### Possible Solution(s)

_No response_

### Notes

_No response_

### Scarb Version

```text

```

wawel37:


tests/e2e/completions/dot.rs
```diff
@@ -978,3 +978,295 @@ fn with_nested_binary_expressions_and_caret_after_with_macro() {
     insert_text = "make(${1:a})"
     "#);
 }
+
+// Regression tests for https://github.com/software-mansion/cairols/issues/1314
```
integraledelebesgue: There's no need to reference this issue here since this PR solves it

tests/e2e/completions/path.rs
```diff
@@ -887,3 +894,418 @@ fn trait_name_suffix_from_other_module() {
     """]
     "#);
 }
+
+// Regression tests for https://github.com/software-mansion/cairols/issues/1314
```
integraledelebesgue: Vide supra

tests/e2e/completions/path.rs
```diff
@@ -887,3 +894,418 @@ fn trait_name_suffix_from_other_module() {
...
+    """
+
+    [[completions]]
+    completion_label = "my_func(...)"
+    completion_label_type_info = "fn() -> u32"
+    insert_text = "my_func()"
+    "#);
+}
+
+fn sorting_dep_fixture() -> Fixture {
```
integraledelebesgue: [`e2e::completions`](https://github.com/software-mansion/cairols/blob/af0e0b92dc40b2769ed48f135a69059d6c2133e8/tests/e2e/completions/mod.rs) module already contains some fixtures. You can move this one there to keep the actual test suite clean

tests/e2e/completions/path.rs
```diff
@@ -887,3 +894,418 @@ fn trait_name_suffix_from_other_module() {
...
+        ")
+    }
+}
+
+#[test]
+fn pub_impl_generic_method_empty_prefix() {
+    test_transform_plain!(Completion, sorting_dep_fixture(), "
+    use alexandria_sorting::MergeSort;
+
+    #[test]
```
integraledelebesgue: Is there any reason to use the `#[test]` attribute here? The transform is plain, so no proc-macro processes it. You can leverage `test_transform_with_macros!` to fully test the behaviour inside tests

tests/e2e/support/transform.rs
```diff
@@ -19,6 +21,10 @@ pub fn conduct_transformation<T: Transformer>(
     let mut fixture = initial_fixture;
     fixture.add_file(T::main_file(), cairo.clone());
 
+    // If the fixture already provides a Scarb.toml, don't add cairo_project.toml — it would
+    // take precedence over Scarb.toml and hide the dependency crates from the LS.
+    let has_scarb_toml = fixture.files().iter().any(|f| f == Path::new("Scarb.toml"));
+
```
integraledelebesgue, line 27: nit: you can move these changes to a separate PR if you want because they're not directly related to this PR

tests/e2e/completions/dot.rs
wawel37: Done

tests/e2e/completions/path.rs
wawel37: Done

tests/e2e/completions/path.rs
wawel37: Done

tests/e2e/completions/path.rs
wawel37: Removed that, i don't remember now why was it there to begin with tbh xd

tests/e2e/support/transform.rs
wawel37: its a small change so i think i will leave it as it is here

integraledelebesgue (approved):

PR #1374, opened 2026-07-24
Add declarative-macro semantic highlighting tests

integraledelebesgue:

PR #1373, opened 2026-07-24
Add proc-macro semantic highlighting tests

integraledelebesgue:
Closes: #1369 

Bugs noted in those tests are fixed later in the stack.

PR #1372, opened 2026-07-24
Migrate semantic tokens tests to `Transformer` infrastructure

integraledelebesgue:
Implementing `Transformer` allows introducing tests of semantic highlighting inside proc macros in the next PR

PR #1375, opened 2026-07-24
Fix identifier resolution in macro-generated content

### LINKED ISSUE #825: Implement a proper semantic highlighting for code inside declarative macros
At the moment, content of inline macro invocations is highlighted identically to the macro itself.
We cannot access the semantics of `TokenTree`s' elements because the compiler does not expose them. We need to wait until a proper API is implemented in the compiler.

integraledelebesgue:
Inline-macro arguments live in a token tree whose virtual expansion file has no
lookup item of its own, so `find_lookup_item` returned `None` and content was
painted as the macro. Collect lookup items from the parent files the resultant
maps to, and look patterns up by the resultant's pointer. This colors variables,
functions, types, enums and paths inside inline, attribute and proc macros.


integraledelebesgue:

- #1372
    - #1373
        - #1374
            - #1375 ◀
                - #1376
                    - #1377
                        - #1378

src/ide/semantic_highlighting/token_kind.rs
```diff
@@ -272,23 +273,23 @@ impl SemanticTokenKind {
 
     /// Arguments:
     /// - `db`: The database to use for lookup.
-    /// - `expr_path_ptr`: The expression path pointer to use for the semantic lookup.
+    /// - `resultant`: The resultant syntax node.
     /// - `lookup_item_id`: The lookup item ID to use for the semantic lookup.
     ///   Returns
     /// - `Some(SemanticTokenKind)` if a corresponding semantic token kind is found, otherwise returns `None`.
     fn from_expr_path(
```
piotmag769, line 280: I have a question - why do we even do this? I checked and this heuristic was introduced by Shahar btw xD Wouldn't all that is handled here be resolved before we enter that - [here](https://github.com/software-mansion/cairols/blob/aa83ff4dc314f52756c696158684e5b41e758ecc/src/ide/semantic_highlighting/token_kind.rs#L240)? Or maybe this is for cases where the resolver fails (e.g. during writing) and we want to color stuff as variable in that case?

piotmag769 (approved):

wawel37 (approved):

src/ide/semantic_highlighting/token_kind.rs
integraledelebesgue: https://github.com/software-mansion/cairols/issues/1385

PR #1376, opened 2026-07-24
Highlight the bang of nested inline macros as a macro

### LINKED ISSUE #825: Implement a proper semantic highlighting for code inside declarative macros

integraledelebesgue:
A nested invocation like `array![array![x]]` is not parsed as an `ExprInlineMacro`;
its inner `!` lives in the outer macro's token tree, so it was highlighted as an
operator. Detect it structurally: a `!` preceded by an identifier and followed by a
bracketed token tree.

### DIFF
--- a/src/ide/semantic_highlighting/token_kind.rs
+++ b/src/ide/semantic_highlighting/token_kind.rs
@@ -140,7 +140,7 @@ impl SemanticTokenKind {
                 if matches!(
                     grandparent_kind,
                     Some(SyntaxKind::ExprInlineMacro | SyntaxKind::ItemInlineMacro)
-                ) =>
+                ) || is_token_tree_macro_bang(db, node) =>
             {
                 Some(SemanticTokenKind::InlineMacro)
             }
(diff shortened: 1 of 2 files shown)


integraledelebesgue:

- #1372
    - #1373
        - #1374
            - #1375
                - #1376 ◀
                    - #1377
                        - #1378

src/ide/semantic_highlighting/token_kind.rs
```diff
@@ -309,6 +309,43 @@ fn find_closest_terminal_ancestor_or_self<'db>(
...
+    }
+    let Some(list) = leaf.parent(db) else {
+        return false;
+    };
+
+    let children = list.get_children(db);
+    let Some(position) = children.iter().position(|child| child.offset(db) == leaf.offset(db))
+    else {
+        return false;
+    };
```
piotmag769, line 333: [nit] both returns are unreachable I guess, but better safe than sorry

piotmag769 (approved):

wawel37 (approved):

PR #1392, opened 2026-09-01
Fix semantic highlighting for locals shadowing crate names

### LINKED ISSUE #1302: Incorrect semantic highlighting for local var with the same name as crate name
`snforge_std` is highlighted as crate while it is a local variable here. Yes, this is a valid cairo code that compiles and executes successfully.
```cairo
#[test]
fn test() {
    let snforge_std = 10;
    assert(snforge_std == 10, '');
}
```

zakazaka95:
## Summary
- prioritize function-local pattern lookup before crate/module resolution during semantic highlighting
- add regression coverage for a local `snforge_std` binding shadowing the crate name

This keeps the local declaration and its use classified as variables instead of namespaces.

## Verification
- added an end-to-end semantic-token regression test
- `git diff --check`
- the Rust test suite was not run locally because this environment has no Rust toolchain; CI should execute the new regression

Closes #1302

### DIFF
--- a/tests/e2e/semantic_tokens/complex.rs
+++ b/tests/e2e/semantic_tokens/complex.rs
@@ -60,6 +60,23 @@ fn complex() {
 }
 
 #[test]
+fn local_variable_shadows_crate_name() {
+    test_transform_plain!(SemanticTokens, r#"
+    #[test]
+    fn test() {
+        let snforge_std = 10;
+        assert(snforge_std == 10, '');
+    }
+    "#, @r#"
+    #[<token=decorator>test</token>]
+    <token=keyword>fn</token> <token=function>test</token>() {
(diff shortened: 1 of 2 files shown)


Arcticae (approved):
LGTM

tests/e2e/semantic_tokens/complex.rs
```diff
@@ -59,6 +59,23 @@ fn complex() {
     ")
 }
 
+#[test]
+fn local_variable_shadows_crate_name() {
+    test_transform_plain!(SemanticTokens, r#"
+    #[test]
+    fn test() {
+        let snforge_std = 10;
```
piotmag769, line 67: This is not declared as a dependency in this project

PR #1393, opened 2026-09-02
Fix stuck analysis spinner

### LINKED ISSUE #1365: Investigate spinner activating and not de-activating after swap
### Problem

After longer period of inactivity cairo spinner shows up and doesn't disappear

Image

### Steps

1. Launch project
2. Wait a few minutes (inactive swap time) without activity 
3. See spinner not deactivating

### Possible Solution(s)

_No response_

### Notes

_No response_

### Scarb Version

```text
2.19.3
```

hakiers:
Closes #1365

## Problem

After a few minutes of inactivity the analysis spinner appears and never goes
away. It only clears once the user types something.

## Cause

`swap_on_inactivity` always sends `AnalysisEvent::DatabaseSwap`, which turns the
spinner **on**. The spinner is turned **off** only when a diagnostics tick
finishes, and ticks are started by `refresh_diagnostics` - a `sync_mut_task`
hook that the scheduler runs only if the task changed the database revision.

A swap *replaces* the database, so the before/after revisions come from
unrelated Salsa runtimes. When nothing has been written since the previous swap,
`migrate_to_fresh_database` rebuilds an identical database with an identical
revision, the gate evaluates to `false`, and the hooks - including the
diagnostics refresh - are skipped. The spinner is switched on unconditionally
while the only thing that can switch it off sits behind a gate the swap fails.

This is why the *second* consecutive idle swap is the one that wedges it: the
first follows real edits, so its revision does differ.

### DIFF
--- a/src/lang/db/swapper.rs
+++ b/src/lang/db/swapper.rs
@@ -50,8 +50,8 @@ pub struct AnalysisDatabaseSwapper {
     mutations_since_last_replace: u64,
     db_replace_min_mutations: u64,
     analysis_event_sender: Sender<AnalysisEvent>,
-    /// Revision of the database produced by the last swap.
-    ///
+    /// Revision of the database produced by the last swap
+    /// (i.e.. the revision of the freshly swapped database).
     /// Used to tell whether anything has actually been written since then, which decides both
     /// whether a new swap can reclaim anything and whether the scheduler will run the
     /// `sync_mut_task` hooks for it.
@@ -73,7 +73,12 @@ impl AnalysisDatabaseSwapper {
     }
 
     /// Swaps the database, triggered by the inactivity monitor.
-    pub fn swap_on_inactivity(&mut self, db: &mut AnalysisDatabase, open_files: &HashSet<Url>) {
+    /// Swap will only be performed if the database has been mutated since the last swap.
+    pub fn maybe_swap_on_inactivity(
+        &mut self,
+        db: &mut AnalysisDatabase,
+        open_files: &HashSet<Url>,
+    ) {
         if self.last_swap_revision == Some(current_revision(db)) {
             trace!("skipping inactivity swap: nothing was written since the last swap");
             return;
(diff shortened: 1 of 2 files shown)


maciektr (approved):

src/lang/db/swapper.rs
```diff
@@ -56,21 +64,28 @@ impl AnalysisDatabaseSwapper {
...
         }
     }
 
     pub fn register_mutation(&mut self) {
         self.mutations_since_last_replace += 1;
     }
 
-    /// Swaps the database unconditionally, triggered by the inactivity monitor.
+    /// Swaps the database, triggered by the inactivity monitor.
     pub fn swap_on_inactivity(&mut self, db: &mut AnalysisDatabase, open_files: &HashSet<Url>) {
```
piotmag769:
```suggestion
    pub fn maybe_swap_on_inactivity(&mut self, db: &mut AnalysisDatabase, open_files: &HashSet<Url>) {
```
To keep the convention

src/lang/db/swapper.rs
```diff
@@ -48,6 +50,12 @@ pub struct AnalysisDatabaseSwapper {
     mutations_since_last_replace: u64,
     db_replace_min_mutations: u64,
     analysis_event_sender: Sender<AnalysisEvent>,
+    /// Revision of the database produced by the last swap.
```
piotmag769:
```suggestion
    /// Revision of the database produced by the last swap (i.e.. revision of the freshly swapped database).
```

src/lang/db/swapper.rs
```diff
@@ -56,21 +64,28 @@ impl AnalysisDatabaseSwapper {
...
+            last_swap_revision: None,
         }
     }
 
     pub fn register_mutation(&mut self) {
         self.mutations_since_last_replace += 1;
     }
 
-    /// Swaps the database unconditionally, triggered by the inactivity monitor.
+    /// Swaps the database, triggered by the inactivity monitor.
```
piotmag769, line 75: Comment when the swap does not happen

piotmag769 (approved):
Good catch

Issue #1293, opened 2026-05-05
Codelens for displaying Gas cost for given function


wawel37:
As title states, we want to display gas price for a given function inside the codelens. It might require profiler integration but not so sure about that. Also can be discussed within team.

PR #1395, opened 2026-09-08
Add gas code lens 

hakiers:
- Adds a new GasCodeLens alongside the existing "▶ Run test(s)" lens, offered on test functions in crates that depend on snforge_std.

This is milestone 1 (tracked in #1293): the lens is wired up and appears in the right place, but execute() is currently a no-op — no gas computation. Subsequent PRs will add the actual snforge measurement.

What's included:
- New GasTarget enum (src/ide/code_lens/gas.rs), derived from TestCodeLensInternal, designed to extend to #[executable] functions later without restructuring
- GasCodeLens/GasCodeLensInternal, following the existing Debugger/Executable lens patterns
- Wired into LSCodeLens and calculate_code_lens

Not included yet (follow-ups): actual snforge execution + gas parsing.

Screenshot 2026-09-08 at 13 59 17


src/ide/code_lens/gas.rs
```diff
@@ -0,0 +1,99 @@
...
+use crate::ide::code_lens::{CodeLens, CodeLensInterface, CodeLensInternal, make_lens_args};
+use crate::lang::db::AnalysisDatabase;
+use crate::lang::lsp::LsProtoGroup;
+use crate::server::client::Notifier;
+use crate::server::commands::ServerCommand;
+use crate::state::State;
+
+#[derive(PartialEq, Clone, Debug)]
+pub enum GasTarget {
+    Test { full_path: String, is_on_mod: bool },
```
piotmag769: Do we want to have a lens that runs gas calculations for all tests inside a module? (we can, just asking if it was discussed)

hakiers: Yes. I discussed it.

piotmag769 (changes requested):
The code is fine but I don't agree with the summarising of gas from multiple that is presented in the next PR - and the idea to run gas lens on modules is introduced in this PR already

src/ide/code_lens/gas.rs
```diff
@@ -0,0 +1,99 @@
...
+    // TODO(hakiers) Calculate gas for #[executable] functions
```
wawel37: link the issue plz

wawel37 (approved):
Looks good imho, lets stay with the concept of not doing it for whole test modules

src/ide/code_lens/gas.rs
piotmag769: You don't need `is_on_mod` if we don't want to have these lenses - just filter out the tests lenses with `is_on_mod: true`

hakiers: I have changed it in the next pr: https://github.com/software-mansion/cairols/pull/1396/changes#diff-ce01ab7196f18f5711e3ed9581bb2cd89a312c027380f82f84866d08f832f2b7R23

piotmag769: Please change it in this one - there is no need to introduce changes that you will revert in the next PR. This creates unnecessary noise

piotmag769 (approved):

PR #1396, opened 2026-09-09
Add gas lens logic

hakiers:
Adds a new CodeLens, "⛽ Calculate Gas", shown alongside the existing "▶ Run test(s)" lens on test functions in crates that depend on snforge_std. Tracked under #1293.

What it does: clicking the lens spawns snforge test <path> --exact on a detached background thread, parses the L2 gas usage out of its output, and reports the result via a ShowMessage notification.
Screenshot 2026-09-11 at 10 13 41
Screenshot 2026-09-11 at 10 13 26


src/ide/code_lens/gas.rs
```diff
@@ -97,3 +101,100 @@ pub fn get_gas_code_lenses(
...
+    };
+    let command = path.snforge_command();
+
+    let file_path = file_url.to_file_path().ok()?;
+    let cwd = state.project_controller.configs_registry().manifest_dir_for_file(&file_path)?;
+
+    // Gas policy:
+    // If the test (or ant test in the module run) fails, we show an error instead of a gas number
+    // If the test is a fuzzer, we report the max L2 gas across all its runs.
+    // If the lens targets a module, we report the sum of L2 gas across all tests in it.
```
piotmag769: I don't think the information is useful to the user. Usually you want to test gas usage of a single execution flow (so one test)

src/ide/code_lens/gas.rs
```diff
@@ -97,3 +101,100 @@ pub fn get_gas_code_lenses(
...
+    std::thread::spawn(move || {
```
piotmag769, line 125: Explain in the comment why it is detached (not joined)

Cargo.toml
```diff
@@ -67,6 +67,7 @@ lsp-server = "0.10.0"
 lsp-types = "=0.95.1"
 memchr = "2.8.1"
 mimalloc = "0.1.52"
+regex = "1.12.3"
```
piotmag769, line 70: We can remove it from dev-dependencies then, right?

piotmag769 (changes requested):
Update the logic of e2e tests (currently calculate gas is skipped there)

hakiers:
> Update the logic of e2e tests (currently calculate gas is skipped there)

https://github.com/software-mansion/cairols/pull/1399

src/ide/code_lens/gas.rs
```diff
@@ -97,3 +101,98 @@ pub fn get_gas_code_lenses(
...
+    } else {
+        TestFullQualifiedPath::Function(full_path.clone())
+    };
+    let command = path.snforge_command();
+
+    let file_path = file_url.to_file_path().ok()?;
+    let cwd = state.project_controller.configs_registry().manifest_dir_for_file(&file_path)?;
+
+    // Running `snforge` can take several seconds, so we let it run in the
+    // background and report the result later via `ShowMessage`z instead of blocking
```
piotmag769:
```suggestion
    // background and report the result later via `ShowMessage` instead of blocking.
```

src/ide/code_lens/gas.rs
```diff
@@ -97,3 +101,98 @@ pub fn get_gas_code_lenses(
...
+    std::thread::spawn(move || {
+        let mut parts = command.split_whitespace();
+        let program = parts.next().expect("command should not be empty");
+        let output = std::process::Command::new(program).args(parts).current_dir(&cwd).output();
+
+        let Ok(output) = output else {
+            let message = "Failed to run snforge".to_string();
```
piotmag769: I would pass the error or the command back to the user so they can see for themselves what failed

src/ide/code_lens/gas.rs
```diff
@@ -97,3 +104,112 @@ pub fn get_gas_code_lenses(
...
+
+        // All-or-nothing: if any test in the batch failed, show error
+        if !output.status.success() {
+            let message = "Test failed or gas disabled".to_string();
+            notifier.notify::<ShowMessage>(ShowMessageParams { typ: MessageType::ERROR, message });
+            return;
+        }
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        let message = match is_fuzzer {
```
piotmag769: Use `if`

src/ide/code_lens/gas.rs
```diff
@@ -96,3 +104,109 @@ pub fn get_gas_code_lenses(
...
+
+        let Ok(output) = output else {
+            let message = format!("Failed to run snforge: {}", output.unwrap_err());
+            notifier.notify::<ShowMessage>(ShowMessageParams { typ: MessageType::ERROR, message });
+            return;
+        };
+
+        // All-or-nothing: if any test in the batch failed, show error
+        if !output.status.success() {
+            let message = "Test failed or gas disabled".to_string();
```
piotmag769: Wdym by "gas disabled"? Also from users POV I have no idea which command failed - I don't know what to rerun to check the failure.

src/ide/code_lens/gas.rs
```diff
@@ -96,3 +104,109 @@ pub fn get_gas_code_lenses(
...
+                L2 Gas: ~{gas} \n
+                Test: {full_path}
+            "
+            )
+        } else {
+            let (max, min, mean) = parse_fuzzer_test_l2_gas(&stdout).unwrap_or_default();
+            format!(
+                "
+                L2 Gas: max: ~{max}, min: ~{min}, mean: ~{mean} \n
+                Test: {full_path}
```
piotmag769: Are the spaces before newlines necessary (maybe they are - just asking)?

src/ide/code_lens/gas.rs
```diff
@@ -96,3 +104,109 @@ pub fn get_gas_code_lenses(
...
+    fn test_parse_fuzzer_test_l2_gas() {
+        let stdout = indoc::indoc! {r#"
+            [PASS] playground::tests::test_one (runs: 256, (l1_gas: {max: ~0, min: ~0, mean: ~0, std deviation: ~0}, l1_data_gas: {max: ~0, min: ~0, mean: ~0, std deviation: ~0}, l2_gas: {max: ~54300, min: ~42660, mean: ~51686, std deviation: ~3360}))
+        "#};
+
+        assert_eq!(parse_fuzzer_test_l2_gas(stdout).unwrap_or_default(), (54300, 42660, 51686));
+    }
+
+    #[test]
+    fn return_zero_when_no_gas_info() {
```
piotmag769: It does not return 0

src/ide/code_lens/gas.rs
hakiers: By gas disabled I mean user can set `[cairo] enable-gas = false` in `Scarb.toml`, causing the test to fail.

src/ide/code_lens/gas.rs
hakiers: Cargo fmt does that

src/ide/code_lens/gas.rs
```diff
@@ -96,3 +104,109 @@ pub fn get_gas_code_lenses(
...
+            return;
+        }
+
+        let message = if is_fuzzer {
+            let (max, min, mean) = parse_fuzzer_test_l2_gas(&stdout).unwrap_or_default();
+            format!(
+                "
+                L2 Gas: max: ~{max}, min: ~{min}, mean: ~{mean}\n
+                Test: {full_path}
+            "
```
wawel37, line 164: why do we keep the newlines here, when the message in vsc appears in one line only?

src/ide/code_lens/gas.rs
```diff
@@ -96,3 +104,109 @@ pub fn get_gas_code_lenses(
...
+            )
+        } else {
+            let gas = parse_test_l2_gas(&stdout).unwrap_or_default();
```
wawel37: if its the stdout out of forge binary, we should rather `.expect` some value rather than just `unwrap_or_default`

src/ide/code_lens/gas.rs
```diff
@@ -30,9 +31,12 @@ pub struct GasCodeLens {
 }
 
 impl CodeLensInterface for GasCodeLens {
-    fn execute(&self, _file_url: Url, _state: &State, _notifier: &Notifier) -> Option<()> {
-        trace!("Execute Calculate Gas");
-        None
+    fn execute(&self, file_url: Url, state: &State, notifier: &Notifier) -> Option<()> {
+        match &self.target {
+            GasTarget::Test { full_path, is_fuzzer } => {
+                test_target_execute(full_path.clone(), *is_fuzzer, file_url, state, notifier)
```
wawel37: i think we should re-evaluate the full_path same way as the test codelenses and debugger code lenses do using `get_full_path_and_module_id`, to make sure its the newest known version of the node path at the time

wawel37 (changes requested):

piotmag769 (approved):

wawel37 (approved):

PR #1408, opened 2026-09-16
Show partial parameter inlay hints on arity mismatch 

### LINKED ISSUE #1248: Make parameter inlay hints work on arity mismatch
Follow-up to #581 

Currently, inlay hints work only if the number of params at call sit is correct:
https://github.com/DelevoXDG/cairols/blob/b1872e5485d51548f6ba2d40cba1d07861a0cdcb/tests/e2e/inlay_hints/params.rs#L209-L209

However, we should aim to display partial inlay hints even if the number of args at call site does not match function arity. 

```diff
#[test]
fn arity_overflow_no_hints() {
    test_transform!(inlay_hint, r#"
    fn foo(a: felt252, b: felt252) -> felt252 { a + b }

    fn main() {
        <sel>foo(1, 2, 3);</sel>
    }
    "#, @r#"
    fn foo(a: felt252, b: felt252) -> felt252 { a + b }

    fn main() {
-        foo (1, 2, 3);
+        foo(a: 1, b: 2, 3);
    }
    "#)
}

#[test]
fn arity_underflow_no_hints() {
    test_transform!(inlay_hint, r#"
    fn foo(a: felt252, b: felt252) -> felt252 { a + b }

    fn main() {
        <sel>foo(1);</sel>
    }
    "#, @r#"
    fn foo(a: felt252, b: felt252) -> felt252 { a + b }

    fn main() {
-       foo(1);
+       foo(a: 1);
    }
    "#)
}
```

hakiers:
Description
Closes #1248

Parameter inlay hints are now shown even when the number of arguments doesn't match the function arity,
e.g. `foo(1)` → `foo(a: 1)`.

While working on this, I found another bug: `self` was always skipped, so calls like `Impl::foo(x, 1)`
got shifted hints (`a: x`). Now `self` is skipped only for method calls (`x.foo(1)`).

---

## Before:
Untitled design (1)

## After:
Untitled design (2)

---


tests/e2e/inlay_hints/params.rs
piotmag769, line 230: Change fixed test names

piotmag769 (approved):
Gj

wawel37 (approved):
Looks good

